PR #86, opened 2024-04-03
fix: reopening About window

azfoo:
- About window was not closed properly and could not be reopened after
  closing as there was no clsoe request sent to remove it from the list
  of already opened windows
About window now posts request to close the window, so that the window is properly removed from the list of opened windows and can be successfully reopened.


tilia/ui/windows/about.py
```diff
@@ -1,8 +1,10 @@
 from PyQt6.QtCore import Qt
+from PyQt6.QtGui import QCloseEvent
```
FelipeDefensor, line 2: Is this needed? I don't see it used anywhere.

azfoo: oops... good catch! I've removed it

PR #90, opened 2024-04-04
refactor subscriptions from __init__ 

### LINKED ISSUE #89: Remove create class method from TimelineUIElement and TimelineComponent
The subclasses of those classes use to realy on a `create` class method for instancing objects. I figured that was no longer needed and removed some of them. I'm not sure I removed all and I am also not sure if any code still relies on  `TimelineUIElement.create()`.

azfoo:
Resolves #89 

Moved all `listen` and `serve` calls to _setup_subscriptions.


FelipeDefensor (commented):
Perfect refactor. I see you named the function `_setup_subscriptions` instead of `_setup_requests`, and I can see why. The idea of naming the pubsub module `requests` was that it's perhaps more like a client/server, because it can also handle "GET" events, that is, events that respond with a payload. That is why I called one of the function that sends events `post` instead of `fire`. I was not so sure about that naming choice at the time, though, as not all aspects of a client/server are there. What are your thoughts on this?

azfoo:
I like what the GET/POST model is trying to get at! It's a naming system that points to an existing idea - a good reference for people to consider when trying to understand the code.

FelipeDefensor:
Ok, so let's stick with the name `requests` for now. Then, you need to rename the `_setup_subscriptions` methods to `_setup_requests` and we're done with this PR.

tilia/timelines/collection/collection.py
FelipeDefensor, line 1: Nice catch.

FelipeDefensor:
Ok. So I'm still away from my coding setup, I will wait until I get that back to make the actual review, but everything seems fine, apart from the fact that the commit history is a little confusing. I can fix that later, no problem.

FelipeDefensor:
So, I finally got around to reviewing this. Sorry for the huge delay. Everything seems fine. The only thing I did was squashing the four rename commits into a single one.

PR #106, opened 2024-05-24
Fix Inspector

### LINKED ISSUE #103: Inspect - dialog does not shrink vertically if it was previously inspecting an element with more rows

azfoo:
Closes #103


azfoo:
Ooops. That should fix it...

FelipeDefensor (approved):

PR #110, opened 2024-05-28
Metadata window width

### LINKED ISSUE #109: Edit metadata window is huge when path is long

azfoo:
Closes #109.

### DIFF
--- a/tilia/settings.py
+++ b/tilia/settings.py
@@ -12,6 +12,9 @@ DEFAULT_SETTINGS = {
         "window_y": 10,
         "timeline_bg": "#EEE",
     },
+    "metadata": {
+        "window-width": 400
+    },
     "auto-save": {"max_saved_files": 100, "interval": 300},
     "media_metadata": {
         "default_fields": [
--- a/tilia/settings.py
+++ b/tilia/settings.py
@@ -12,9 +12,6 @@ DEFAULT_SETTINGS = {
         "window_y": 10,
         "timeline_bg": "#EEE",
     },
-    "metadata": {
-        "window-width": 400
-    },
     "auto-save": {"max_saved_files": 100, "interval": 300},
     "media_metadata": {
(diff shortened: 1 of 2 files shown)


tilia/settings.py
```diff
@@ -12,6 +12,9 @@
         "window_y": 10,
         "timeline_bg": "#EEE",
     },
+    "metadata": {
+        "window-width": 400
+    },
```
FelipeDefensor, line 17: We already have a "media_metadata" section, right below. Could we use that?

tilia/ui/windows/metadata.py
```diff
@@ -32,6 +33,7 @@ class MediaMetadataWindow(QDialog):
     def __init__(self):
         super().__init__()
         self.setWindowTitle("Metadata")
+        self.setFixedWidth(settings.get("metadata", "window-width"))
```
FelipeDefensor, line 36: This is good, but if there's no reason not to we should go with `self.resize(settings.get("metadata", "window-width"), self.height())` to keep the window resizable.

FelipeDefensor (changes requested):

tilia/settings.py
azfoo: Good call

tilia/ui/windows/metadata.py
azfoo: Fixed Width:
image

Regular Resize:
image

Word wrapping is not very clever - someone with a silly file name could still accidentally make the window very long. But if that's compromisable, I agree that regular resizing provides more flexibility.

FelipeDefensor: Oh, I see. I still slightly prefer the resize option. In the future, we could add some processing to the string to handle those silly cases.

FelipeDefensor (approved):

PR #113, opened 2024-05-29
Improve user interaction with slider timeline

azfoo:
Slider UI shows the potential for being dragged to seek time.
Added open hand cursor to slider trough that becomes closed when clicked on.

### DIFF
--- a/tilia/settings.py
+++ b/tilia/settings.py
@@ -34,8 +34,8 @@ DEFAULT_SETTINGS = {
         "default_height": 40,
         "trough_radius": 5,
         "trough_color": "#FF0000",
-        "line_color": "#000000",
-        "line_weight": 3,
+        "line_color": "#cccccc",
+        "line_weight": 5,
     },
     "beat_timeline": {"display_measure_periodicity": 4, "default_height": 35},
     "hierarchy_timeline": {
--- a/tilia/ui/timelines/base/timeline.py
+++ b/tilia/ui/timelines/base/timeline.py
@@ -238,7 +238,7 @@ class TimelineUI(ABC):
             elm.on_right_click(x, y, item)
 
     def on_left_click(
-        self, item: QGraphicsItem, modifier: ModifierEnum, double: bool
+        self, item: QGraphicsItem, modifier: ModifierEnum, double: bool, x: int, y: int
     ) -> None:
         clicked_elements = self.get_item_owner(item)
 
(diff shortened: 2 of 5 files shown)


FelipeDefensor (changes requested):
Giving visual feedback for the interactivity of the trough is good and you got the mixin right, but I think the open hand cursor looks bad. Its too big compared to other UI elements and it has an outdated feel to it. The pointing hand cursor looks much better, in my opinion, and could be used both for the dragging and non-dragging state. What do you think?

FelipeDefensor:
On 41ab0b9 I also added some interactivity to the line, which is something that I've seen users expect when trying out TiLiA.

azfoo:
> The pointing hand cursor looks much better, in my opinion, and could be used both for the dragging and non-dragging state. What do you think?

I think the cursors should show standardised interaction across the different timelines. Perhaps the pointing hand if there is something inspectable? In this case, perhaps the horizontal arrows might be better.

FelipeDefensor:
> > The pointing hand cursor looks much better, in my opinion, and could be used both for the dragging and non-dragging state. What do you think?
> 
> I think the cursors should show standardised interaction across the different timelines. Perhaps the pointing hand if there is something inspectable? In this case, perhaps the horizontal arrows might be better.

Fair point. The horizontal arrows look a bit out of place for me here, but it might be just that I'm not used to them yet. Let's go with them.

azfoo:
I'll go ahead and check all of the cursors then. 
On that point, would it be fair to make the hierarchy posts have the same horizontal arrows for cursors? It would make it more apparent that the blocks are sizeable that way too.

FelipeDefensor:
> > > The pointing hand cursor looks much better, in my opinion, and could be used both for the dragging and non-dragging state. What do you think?
> > 
> > 
> > I think the cursors should show standardised interaction across the different timelines. Perhaps the pointing hand if there is something inspectable? In this case, perhaps the horizontal arrows might be better.
> 
> Fair point. The horizontal arrows look a bit out of place for me here, but it might be just that I'm not used to them yet. Let's go with them.

Ok, so I think I get why the horizontal arrow cursor bothers. It's because we're used to it indicating to *edges* that are draggable, which is not the case here. I say we go with this for now and try to get some user feedback, but let's consider modifying the trough drawing when its hovered instead of changing the cursor as an alternative.

FelipeDefensor:
Also, let's squash the first three commits.

FelipeDefensor (approved):

PR #116, opened 2024-05-29
Metadata fields handling

### LINKED ISSUE #112: Editing Metadata Fields
Removing x original metadata fields, then reopening the edit window to remove y causes app to crash. App is attempting to remove x from the list again, even though it has already done so. 
Removing fields after adding new fields does not cause this issue.

azfoo:
Closes #112, #114, #115.


tilia/ui/windows/metadata.py
```diff
@@ -131,10 +131,5 @@ def closeEvent(self, event, **kwargs):
...
 
     def update_metadata_fields(self, new_fields: list[str]):
-        for i, field in enumerate(new_fields):
-            if field not in self.metadata:
-                post(Post.METADATA_ADD_FIELD, field.lower(), i)
-
-        for field in self.metadata:
-            if field not in new_fields + get(Get.MEDIA_METADATA_REQUIRED_FIELDS):
-                post(Post.METADATA_REMOVE_FIELD, field)
+        if not [item for item in self.metadata.keys() if item not in get(Get.MEDIA_METADATA_REQUIRED_FIELDS)] == new_fields:
```
FelipeDefensor, line 134: I get this, but it's a little hard to read. Let's add a comment or separate the creation of the list from the comparison.

FelipeDefensor (changes requested):
Apart from the small readability improvement I mentioned, I suggest adding tests for:
- Ensuring required fields are not duplicated if user also inserts in the edit fields dialog;
- Ensuring the "title" field is always on top;
- Ensuring the edge case of deleting all fields is handled properly.

Once that's done, squash all the commits into a single one and we're good to go.

FelipeDefensor (approved):

PR #118, opened 2024-05-29
set timeline minimum height

### LINKED ISSUE #117: Changing timeline height to less than 10 silently fails

azfoo:
Closes #117.


FelipeDefensor (approved):
I would prefer something that warns the user explicitly about the timeline size limit. If we enter a number lower than 10 in the dialog the "Ok" button is just greyed out, without a clear reason.

Even then, I think this is a clever solution, and good enough for now. We'll need to remember to impose this value limit on other ways of changing timeline height, if they arise in the future, since we're not limiting this on the timeline level.

PR #122, opened 2024-06-01
Fix Inspector again

### LINKED ISSUE #119: Inspect window shouldn't be blank when nothing is selected
It seems like we're clearing all widgets from the inspect window once an element gets deselected, to facilitate repopulating it later. That's fine, but in the case where there's no element selected the window is just blank, which looks pretty bad. Instead, we should display disabled widgets or some text like "No element selected".

azfoo:
closes #119

### DIFF
--- a/tilia/ui/windows/inspect.py
+++ b/tilia/ui/windows/inspect.py
@@ -15,7 +15,10 @@ from PyQt6.QtWidgets import (
     QSizePolicy,
     QSpinBox,
     QComboBox,
+    QStackedWidget,
 )
+
+from PyQt6.QtCore import Qt
 
 from tilia.requests import Post, listen, stop_listening_to_all, post, Get, get
 from tilia.utils import get_tilia_class_string
@@ -46,16 +49,26 @@ class Inspect(QDockWidget):
     def __init__(self) -> None:
         super().__init__()
         self.setWindowTitle("Inspector")
+        self.setMinimumSize(400, 0)
         self._setup_requests()
         self.inspected_objects_stack = []
         self.element_id = ""
-        self.widget = QWidget(self)
-        self.setWidget(self.widget)
-        self.layout = QFormLayout(self.widget)
-        self.widget.setLayout(self.layout)
-
         self.currently_inspected_class = None
         self.field_name_to_widgets = None
+
+        self.stack_widget = QStackedWidget(self)
+        self.setWidget(self.stack_widget)
+
+        self.inspect_widget = QWidget(self.stack_widget)
+        self.inspect_layout = QFormLayout(self.inspect_widget)
+        self.inspect_widget.setLayout(self.inspect_layout)
+
+        self.empty_label = QLabel("<h2 style='color:grey'>No element selected.</h2>")
+        self.empty_label.setAlignment(Qt.AlignmentFlag.AlignCenter)
+
+        self.stack_widget.addWidget(self.inspect_widget)
+        self.stack_widget.addWidget(self.empty_label)
+
         post(Post.WINDOW_INSPECT_OPENED)
 
     def __str__(self):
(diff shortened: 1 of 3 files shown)


tilia/ui/windows/inspect.py
```diff
@@ -97,6 +99,7 @@ def on_element_selected(
         inspector_values: dict[str, str],
         element_id: int,
     ):
+        self.setUpdatesEnabled(False)
```
FelipeDefensor, line 102: Is this supposed to prevent flickering when selecting elements? If so, this is a nice find and should come in handy in other places. It doesn't work for me, however (see video below).


https://github.com/FelipeDefensor/TiLiA/assets/87732202/798dce27-c769-4cf5-9ee2-58793ff5b5a0

tilia/ui/windows/inspect.py
```diff
@@ -56,6 +57,7 @@ def __init__(self) -> None:
 
         self.currently_inspected_class = None
         self.field_name_to_widgets = None
+        self.empty_label = QLabel("No element selected.")
```
FelipeDefensor, line 60: Let's make the font a bit bigger and centralize this. The widget looks too empty as it is.

inspector-no-element-selected

FelipeDefensor (changes requested):

tilia/ui/windows/inspect.py
azfoo: It doesn't work as well as I hoped, but it does flicker a little less.

tilia/ui/windows/inspect.py
azfoo: Not too sure about the center align. See latest push.

FelipeDefensor:
Will take a look at the changes once I finish working on the PDF timeline. In the meantime, I found this:

https://github.com/FelipeDefensor/TiLiA/assets/87732202/3848b1c5-ad38-4953-962d-d0155201712a

azfoo:
> There's an empty space below the PDF inspector if we first open the inspector while it is selected.

I can't tell for sure, but it could be a branch thing. 
There is a notable gap after the last item on the table even on the latest update but I don't see anything that could be causing it.

FelipeDefensor:
> > There's an empty space below the PDF inspector if we first open the inspector while it is selected.
> 
> I can't tell for sure, but it could be a branch thing. There is a notable gap after the last item on the table even on the latest update but I don't see anything that could be causing it.

It's a branch thing, sorry. I was (obviously) using the pdf timeline branch when I made the video. I do just get the small gap when selecting beat on beat timelines.

tilia/ui/windows/inspect.py
FelipeDefensor: We should also center it vertically, that will look much better. The problem is that gap in the bottom, once again. I don't want to waste a lot of time here, but I think we should try to remove that, if only to get this looking better. I also think we should go for bold instead of italic and use a more muted color.

We could sidestep this using the pevious strategy of disabling the widgets if there are no elements selected. I think that looks way better than the "No element selected" ever could. Could you perhaps check if the selected stack is empty and if so trigger a disable instead of a clear?

tilia/ui/windows/inspect.py
FelipeDefensor: That's the problem with rebuilding the widget from scratch on every new select/deselect. That may have been why I choose to repopulate it instead, I can't quite remember. Anyways, the flicker is better than the mounting dead space on the bottom.

tilia/ui/windows/inspect.py
azfoo: Should be fixed now

FelipeDefensor (approved):
The stacked widget was a good solution. +1 for also improving the separator.

FelipeDefensor:
@azfoo this is approved. I have been rebasing and merging your PRs myself after approval, but I like the idea of the author doing that better. It gives a last chance to make any changes that might have occured to her between the review request and the approval. Let's start doing that with this PR, ok?

Don't forget that we are on the rebase team, so choose "Rebase and merge" instead of "Create a merge commit".

azfoo:
I think I did that right...
Will do the rebasing and merging myself in the future after your review approval.

PR #127, opened 2024-06-05
Fix Crash Handler Sizing

### LINKED ISSUE #126: Crash handler is higher than monitor size and very hard to resize when traceback is long
We should use a textbox with a max height and a scrollbar to display the traceback instead of the current one.

azfoo:
Closes #126


FelipeDefensor (approved):
This works perfectly and there's no need to change anything, but I wonder if just using a `QTextEdit` wouldn't have been easier? As per the docs (https://doc.qt.io/qt-6/qtextedit.html):

image

PR #140, opened 2024-06-15
Migrating from toml to qSettings

### LINKED ISSUE #138: Add Open recent... menu

azfoo:
- Possibly closes: https://github.com/users/FelipeDefensor/projects/11/views/1?pane=issue&itemId=66679501
- closes #138 
- Also added settings window for easier access to settings

### DIFF
--- a/tilia/settings.py
+++ b/tilia/settings.py
@@ -69,7 +69,7 @@ DEFAULT_SETTINGS = {
         "marker_height": 10,
         "marker_default_color": "#999999",
     },
-    "pdf_timeline": {
+    "PDF_timeline": {
         "default_height": 30,
     },
     "harmony_timeline": {"default_harmony_display_mode": "chord"},
--- a/tilia/settings.py
+++ b/tilia/settings.py
@@ -13,7 +13,7 @@ class SettingsManager(QObject):
             "window_y": 10,
             "timeline_background_color": "#EEE",
         },
-        "auto-save": {"max_number_saved_files": 100, "interval": 300},
+        "auto-save": {"max_stored_files": 100, "interval_(seconds)": 300},
         "media_metadata": {
             "default_fields": [
                 "composer",
@@ -40,13 +40,13 @@ class SettingsManager(QObject):
         },
         "audiowave_timeline": {
             "default_height": 80,
-            "wave_color": "#3399FF",
+            "default_color": "#3399FF",
             "max_divisions": 2500
         },
         "beat_timeline": {"display_measure_periodicity": 4, "default_height": 35},
         "hierarchy_timeline": {
             "default_height": 120,
-            "hierarchy_default_colors": [
+            "default_colors": [
                 "#ff964f",
                 "#68de7c",
                 "#f2d675",
@@ -59,15 +59,15 @@ class SettingsManager(QObject):
                 "#a7aaad",
                 "#4f94d4",
             ],
-            "hierarchy_base_height": 25,
-            "hierarchy_level_height_diff": 25,
-            "hierarchy_marker_height": 10,
+            "base_height": 25,
+            "level_height_diff": 25,
+            "divider_height": 10,
         },
         "marker_timeline": {
             "default_height": 30,
             "marker_width": 8,
             "marker_height": 10,
-            "marker_default_color": "#999999",
+            "default_color": "#999999",
         },
         "PDF_timeline": {
             "default_height": 30,
--- a/tilia/settings.py
+++ b/tilia/settings.py
@@ -1,126 +1,174 @@
-from pathlib import Path
-import tomlkit
+from PyQt6.QtCore import QSettings, QObject
 
-from tilia.utils import open_with_os
+import tilia.constants
 
-DEFAULT_SETTINGS = {
-    "general": {
-        "auto-scroll": False,
-        "window_width": 800,
-        "window_height": 400,
-        "window_x": 20,
-        "window_y": 10,
-        "timeline_bg": "#EEE",
-    },
-    "auto-save": {"max_saved_files": 100, "interval": 300},
-    "media_metadata": {
-        "default_fields": [
-            "composer",
-            "tonality",
-            "time signature",
-            "performer",
-            "performance year",
-            "arranger",
-            "composition year",
-            "recording year",
-            "form",
-            "instrumentation",
-            "genre",
-            "lyrics",
-            "notes",
-        ],
-        "window_width": 400,
-    },
-    "slider_timeline": {
-        "default_height": 40,
-        "trough_radius": 5,
-        "trough_color": "#FF0000",
-        "line_color": "#cccccc",
-        "line_weight": 5,
-    },
-    "audiowave_timeline": {
-        "default_height": 80,
-        "wave_color": "#3399FF",
-        "max_div": 2500
-    },
-    "beat_timeline": {"display_measure_periodicity": 4, "default_height": 35},
-    "hierarchy_timeline": {
-        "default_height": 120,
-        "hierarchy_default_colors": [
-            "#ff964f",
-            "#68de7c",
-            "#f2d675",
-            "#ffabaf",
-            "#dcdcde",
-            "#9ec2e6",
-            "#00ba37",
-            "#dba617",
-            "#f86368",
-            "#a7aaad",
-            "#4f94d4",
-        ],
-        "hierarchy_base_height": 25,
-        "hierarchy_level_height_diff": 25,
-        "hierarchy_marker_height": 10,
-    },
-    "marker_timeline": {
-        "default_height": 30,
-        "marker_width": 8,
-        "marker_height": 10,
-        "marker_default_color": "#999999",
-    },
-    "PDF_timeline": {
-        "default_height": 30,
-    },
-    "harmony_timeline": {"default_harmony_display_mode": "chord"},
-    "dev": {"log_events": False, "log_requests": False, "dev_mode": False},
-}
+class SettingsManager(QObject):
 
-_settings = DEFAULT_SETTINGS
-_settings_path = Path()
+    DEFAULT_SETTINGS = {
+        "general": {
+            "auto-scroll": 'false',
+            "window_width": 800,
+            "window_height": 400,
+            "window_x": 20,
+            "window_y": 10,
+            "timeline_background_color": "#EEE",
+        },
+        "auto-save": {"max_number_saved_files": 100, "interval": 300},
+        "media_metadata": {
+            "default_fields": [
+                "composer",
+                "tonality",
+                "time signature",
+                "performer",
+                "performance year",
+                "arranger",
+                "composition year",
+                "recording year",
+                "form",
+                "instrumentation",
+                "genre",
+                "lyrics",
+            ],
+            "window_width": 400,
+        },
+        "slider_timeline": {
+            "default_height": 40,
+            "trough_radius": 5,
+            "trough_color": "#FF0000",
+            "line_color": "#cccccc",
+            "line_weight": 5,
+        },
+        "audiowave_timeline": {
+            "default_height": 80,
+            "wave_color": "#3399FF",
+            "max_divisions": 2500
+        },
+        "beat_timeline": {"display_measure_periodicity": 4, "default_height": 35},
+        "hierarchy_timeline": {
+            "default_height": 120,
+            "hierarchy_default_colors": [
+                "#ff964f",
+                "#68de7c",
+                "#f2d675",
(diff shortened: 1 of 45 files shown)


azfoo:
TODO:
Reorder settings window
Save geometry and state

tilia/settings.py
```diff
@@ -76,51 +75,56 @@
...
 }
 
-_settings = DEFAULT_SETTINGS
-_settings_path = Path()
+_settings = NotImplemented
 
+def load(reset = False):
+    global _settings
+    _settings = QSettings(tilia.constants.APP_NAME, f"Desktop-v.{tilia.constants.VERSION}")
+    if reset or (not _settings.allKeys() or len(_settings.allKeys()) < sum(len(setting) for setting in DEFAULT_SETTINGS.values())):
```
FelipeDefensor, line 83: I think it's fair to say that the second operand is hard to read. Maybe write a nested function for getting `not _settings.allKeys() or len(_settings.allKeys()) < sum(len(setting)` just so you can add a name to it?

FelipeDefensor: A comment would also do.

tilia/settings.py
```diff
@@ -76,51 +75,56 @@
...
 
-def _set_default_table(table):
-    _settings[table] = DEFAULT_SETTINGS[table]
-    _save()
 
 
-def open_settings_on_os():
-    open_with_os(_settings_path)
+if __name__ == "__main__":
+    load(True)
```
FelipeDefensor, line 158: Now this confuses me. Why would we ever run this file? Was this for testing?

tilia/ui/qtui.py
```diff
@@ -89,17 +89,18 @@ def closeEvent(self, event):
         event.ignore()
 
     def on_close(self):
-        settings.edit("general", "window_width", self.width())
-        settings.edit("general", "window_height", self.height())
-        settings.edit("general", "window_x", self.x())
-        settings.edit("general", "window_y", self.y())
```
FelipeDefensor, line 95: Why on earth have I named that function `edit` in the first place is beyond me. The module is literally called SETtings. Thanks for this improvement.

tilia/ui/qtui.py
```diff
@@ -131,16 +132,17 @@ def _setup_requests(self):
...
             (Post.WINDOW_INSPECT_CLOSE, lambda: self.on_window_close(WindowKind.INSPECT)),
             (Post.WINDOW_INSPECT_CLOSED, lambda: self.on_window_close_done(WindowKind.INSPECT)),
+            (Post.WINDOW_INSPECT_OPEN, lambda: self.on_window_open(WindowKind.INSPECT)),
             (Post.WINDOW_MANAGE_TIMELINES_CLOSE_DONE, lambda: self.on_window_close_done(WindowKind.MANAGE_TIMELINES)),
+            (Post.WINDOW_MANAGE_TIMELINES_OPEN, lambda: self.on_window_open(WindowKind.MANAGE_TIMELINES)),
             (Post.WINDOW_METADATA_CLOSED, lambda: self.on_window_close_done(WindowKind.MEDIA_METADATA)),
-            (Post.WINDOW_ABOUT_CLOSED, lambda: self.on_window_close_done(WindowKind.ABOUT)),
+            (Post.WINDOW_METADATA_OPEN, lambda: self.on_window_open(WindowKind.MEDIA_METADATA)),
+            (Post.WINDOW_SETTINGS_CLOSED, lambda: self.on_window_close_done(WindowKind.SETTINGS)),
+            (Post.WINDOW_SETTINGS_OPEN, lambda: self.on_window_open(WindowKind.SETTINGS)),
```
FelipeDefensor, line 145: We should prefer to order things alphabetically on separate commits. Just a reminder, no need to change anything.

FelipeDefensor (changes requested):
The settings window is a dream come true and the code looks good.

There's one important thing to adress though. The code that uses the settings is not getting notified of changes in those. If we change the timeline background color, for instance, it only gets. There's lot of ways to do this, of course, but I think that a particularly elegant one is to accept and optional callback function as an argument to `settings.get` that gets called whenever that value is set. Feel free to implement this as you wish, though.

Also, we are missing "Apply" and "Cancel" buttons.

tilia/settings.py
azfoo: Mostly for testing, but also if you want to reset everything. Should we keep it?

azfoo:
> There's one important thing to adress though. The code that uses the settings is not getting notified of changes in those. If we change the timeline background color, for instance, it only gets. There's lot of ways to do this, of course, but I think that a particularly elegant one is to accept and optional callback function as an argument to `settings.get` that gets called whenever that value is set. Feel free to implement this as you wish, though.

Will look into that. 

> Also, we are missing "Apply" and "Cancel" buttons.

Will do.

tilia/settings.py
FelipeDefensor: Let's not keep it, the convention is to reserve this idiom for the entry point of the application. If we want a convenient reset let's add a function for that.

FelipeDefensor:
Correcting an incomplete phrase on one of the comments:
> The code that uses the settings is not getting notified of changes in those. If we change the timeline background color settings, for instance, the color only actually changes after restarting the app.

I think you managed to understand without that, though.

azfoo:
Everything in the settings changes on its own now, except for auto-scroll and window geometry. Having those change when settings are meant to be default and universal across all instances of the app didn't make sense.

FelipeDefensor (changes requested):
I haven't finished the review yet, but I think there are some things to work on before I continue.

# Callback pattern desing
Firstly, I like the overall implementation of the callback idea, and it seems to work just fine, good job! There's only one major point I would change. The consumers of the settings are having to call the `link_setting_update` functions themselves. That is not ideal, as it is very easy to forget linking when writing new code that consumes settings. Also, it couples all consumers to the link function, which makes it so that if we had to change it, we would also have to potenttially change code in every consumer.

What I suggest is that the call to `settings.get` already triggers the linking in some way, perhaps by taking a callback as an optional argument.

# Class attribute naming
I see that we need to change a lot of class atttributes that are uppercase, due to them being constants before this implementation. E.g. here:

```
            hierarchy_ui.BASE_HEIGHT = settings.get("hierarchy_timeline", "hierarchy_base_height")
            hierarchy_ui.X_INCREMENT_PER_LVL = settings.get("hierarchy_timeline", "hierarchy_level_height_diff")
            hierarchy_ui.HANDLE_HEIGHT = settings.get("hierarchy_timeline", "hierarchy_marker_height")
```

As they are not constants anymore, the identifiers should be converted to snake case.

# Settings.get cost
Just to make sure, is settings.get expensive? I'm imagining it isn't.

# Errors
I'm getting an error when opening the metadata window.

# Other adjustments
- If that's now too hard to change, the "Dev" section of the setting should be at the bottom.
- "Save" button should read "Apply", in my opinion.
- Some settings seems to appear only occasionally (Dev mode and log requests definitely, and I had the same impression from others). I wasn't able to reproduce this consistently.
- It's not clear whether the auto-save interval is in seconds or minutes. Would it be too incovenient to be able to customize the setting label? we could name it "Interval (seconds)", for instance.
- In the same vein, "Max Number Saved Files" could change to "Max stored files" in the auto-save section.
- Can we add a label to the hierarchy colors indicating which level they refer to?
- In the hierarchy timeline settings, "marker height" should read "divider height" or something like that.
- Should there be a minimum height to the beat timeline to prevent the user from hiding the labels accidentally?
- The save settings dialog should also have a cancel button. Closing the dialog instead of pressing any buttons should have the same effect as using the cancel button.

tilia/settings.py
```diff
@@ -112,6 +121,27 @@ def get(self, group_name: str, setting: str, in_default = True):
...
+                if functions[0]:
+                    functions_to_call.update(functions[1])
+                    functions[0] = False
+
+        functions_to_remove = set()
+        for function in functions_to_call:
+            try:
+                function()
+            except:
+                functions_to_remove.add(function)
```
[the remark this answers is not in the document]
azfoo: Callbacks are functions linked to instances of objects and settings doesn't get informed of deletes that happen, settings might call a deleted object and throw an error. There isn't a way to listen to deletes because that causes circular imports so this seemed to be the only way to remove links to non-existent objects.

azfoo:
> Just to make sure, is settings.get expensive? I'm imagining it isn't.
>
Doesn't seem to be...

> I'm getting an error when opening the metadata window.
> 
my bad. was experimenting with github desktop but that causes all kinds of rebasing problems. should be fixed now.

Regarding names:
will rename them, but just so you know, it'll rename all settings names across the app, which I don't think will be a problem as they are strings anyway.

> * Should there be a minimum height to the beat timeline to prevent the user from hiding the labels accidentally?
For uniformity's sake, there is a minimum of 10 on all height settings (see previous fix), and PDF's height cannot be changed (because that's also a thing that was decided). I'd say that like with PDF, there isn't much of a point in changing the height of this timeline and we could do without showing this setting and also remove the ability to change this timeline's height in other places.

> * The save settings dialog should also have a cancel button. Closing the dialog instead of pressing any buttons should have the same effect as using the cancel button.
To clarify on buttons in general, do we have a cancel and a discard button? Currently, the discard button reverts all settings to when the settings window was first opened and leaves the window open. In the same vein, should the save button also close the window? 

> * Changing the trough radius makes it misalign. If you want to hardcode the radiu instead of dealing with this, I'm fine with that.
That's a problem with my maths - will try to fix that, but if not, it'll be hardcoded.

FelipeDefensor:
> One of the main reasons for designing it this way is that with elements that delete every time a merge/split/redo/undo happens, the number of callbacks stored grows very quickly - which can cause things to become expensive. An optional callback is possible for things outside of timeline UI elements, but I'm not confident that they will continue working on the elements (especially as more components are created).

Good point, but I think we can the first part by moving by making the timeline ui the consumer instead of the elements. It would pass a callback that, when triggered, loops through the elements making the relevant changes.

> Doesn't seem to be [expensive]...

Fine, nothing to comment in that case.

> will rename them, but just so you know, it'll rename all settings names across the app, which I don't think will be a problem as they are strings anyway.

What I am suggesting is decoupling the internal name from the displayed name (again, if that's not too much trouble). In that case, we could keep the current internal name, so no app wide renaming would be needed, right?

> For uniformity's sake, there is a minimum of 10 on all height settings (see previous fix), and PDF's height cannot be changed (because that's also a thing that was decided). I'd say that like with PDF, there isn't much of a point in changing the height of this timeline and we could do without showing this setting and also remove the ability to change this timeline's height in other places.

I agree, let's make the height fixed.

tilia/settings.py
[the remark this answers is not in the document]
FelipeDefensor: Ok, so what about we leverage the request system here? Instead of linking by calling a function from `settings`, the consumers subscribed to the changes using something like `listen(self, Post.SETTINGS_CHANGED)`? That way, when they call `stop_listening_to_all()` (which almost every object needs to do anyway) the callback "link" would be undone. The obvious downside is that any setting that changes might trigger the callback of a lot of objects, but I think the performance is negligible if we move consumers up the design hierarchy (see my comment below about making timeline uis the consumers instead of the elements). Even if that's not the case, we can get around that by dynamically creating "post identifiers", which shouldn't be hard.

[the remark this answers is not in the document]
FelipeDefensor: I know I said that you should go ahead and "implement this however you wanted", so I'm sorry for intruding on this now. I thought the implementation would be more straightforward than it proved to be.

azfoo:
Mostly corrected, except for renaming settings. 
Is there any reason for keeping the internal naming system and what is displayed separately? It seems like extra work trying to decode which setting links to which if we have different naming systems.

FelipeDefensor:
> Mostly corrected, except for renaming settings. Is there any reason for keeping the internal naming system and what is displayed separately? It seems like extra work trying to decode which setting links to which if we have different naming systems.

Should I review, then?

azfoo:
run 
`from tilia.settings import settings
settings.reset_to_default()`
to clear existing settings before running to avoid seeing funny names

azfoo:
did the lazy thing of renaming everything instead. could we perhaps leave the mapping to a later point?

FelipeDefensor:
> did the lazy thing of renaming everything instead. could we perhaps leave the mapping to a later point?

Sure we can. I certainly have postponed a lot of inconvenient coding in this project. You should feel that you can do the same. Let's only keep an eye on each other so the technical debt doesn't get out of hand.

FelipeDefensor (approved):
Squash and merge. I'm very happy about the settings window.

PR #143, opened 2024-06-24
Wheel Events

### LINKED ISSUE #141: Zooming is causing vertical scrolling of timeline area
This is only observable, of course, when the timelines are "taller" than the visible timeline area, which causes a vertical scrollbar to appear.

azfoo:
closes #141 
Fixes accidental scroll on zoom
Added horizontal and vertical zooming through modifiers

### DIFF
--- a/tilia/ui/timelines/collection/collection.py
+++ b/tilia/ui/timelines/collection/collection.py
@@ -327,7 +327,6 @@ class TimelineUIs:
         self.scene.setSceneRect(0, 0, width, self.get_scene_height())
         for tlui in self:
             tlui.set_width(width)
-        self.view.center_on(*self.center_after_zoom)
 
     def update_timeline_ui_ordinal(self):
         self.update_timeline_uis_position()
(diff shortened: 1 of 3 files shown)


azfoo:
It's mostly for you to have a play around and see what you think. 
Vertical scrolling might not be the most useful feature but horizontal scrolling should be handy considering most of the data in the app is along the x-axis. 
Standardised wheel events might be more intuitive for the first-time user, and zooming is still only two clicks away either by the shortcut or the view buttons.

FelipeDefensor:
I'm still getting the extra vertical scrolling. Maybe some Qt default?


https://github.com/TimeLineAnnotator/desktop/assets/87732202/0dcdcc3e-25a4-47ed-8c2d-32020389c386

I like the Shift+Wheel for horizontal scrolling!

I'm still not sure about the horizontal zooming, but good point about it still being available farily easily. I was thinking that eventually we need to add one of those percentage and slider interfaces for zooming that most document viewers have ...

image

... so what about we postpone changing the sortcuts until then?

azfoo:
that should fix it - though it might have accidentally fixed something that was meant for the next version too.

azfoo:
not quite commented out, just stored in a function that is never called. switching over will just involve changing/replacing function names.

FelipeDefensor:
> not quite commented out, just stored in a function that is never called. switching over will just involve changing/replacing function names.

That's even better. Feel free to merge.

PR #147, opened 2024-06-24
fix spacing

### LINKED ISSUE #146: Fix "11" superscript getting cropped on roman numerals
Image

azfoo:
closes #146


FelipeDefensor (approved):
The MusAnalysis font is great, don't get me wrong, but I think we will have to take matters into our own hands eventually to avoid having to resort to this kind of hack. Nice simple fix, though, that's the best we can do for now. Feel free to merge.

azfoo:
it is probably possible to edit the musanalysis font to fix this properly, but we could save that for a later point.

PR #152, opened 2024-07-04
reopening windows

### LINKED ISSUE #148: Can't reopen PDF or video windows if they are closed by the OS
I suggest that we create a "View" menu in the main menu bar where users can toggle the visiblity of those windows and of the Inspect window.

azfoo:
closes #148, #151
make pdf, video and youtube windows reopenable

### DIFF
--- a/tilia/media/player/qtvideo.py
+++ b/tilia/media/player/qtvideo.py
@@ -1,28 +1,19 @@
 from __future__ import annotations
 
-from PyQt6.QtCore import Qt
 from PyQt6.QtMultimediaWidgets import QVideoWidget
 from PyQt6.QtWidgets import QSizePolicy
 
 from .qtplayer import QtPlayer
+from tilia.ui.windows.view_window import ViewWindow
+
 
 class QtVideoPlayer(QtPlayer):
     MEDIA_TYPE = "video"
 
     def __init__(self):
         super().__init__()
-        self.widget = QVideoWidget()
-        self.widget.setSizePolicy(
-            QSizePolicy.Policy.Expanding, QSizePolicy.Policy.Expanding
-        )
+        self.widget = QVideoWindow()
         self.player.setVideoOutput(self.widget)
-        self.widget.setWindowTitle("TiLiA Player")
-        self.widget.setWindowFlags(
-            Qt.WindowType.Window
-            | Qt.WindowType.WindowTitleHint
-            | Qt.WindowType.CustomizeWindowHint
-        )
-        self.widget.resize(800, 600)
 
     def _engine_load_media(self, media_path: str) -> bool:
         result = super()._engine_load_media(media_path)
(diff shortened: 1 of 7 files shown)


tilia/ui/timelines/pdf/timeline.py
```diff
@@ -222,3 +213,42 @@ def update_displayed_page(self, time):
...
+            self.blockSignals(True)
+            self.setVisible(to_show)
+            self.blockSignals(False)
+
+    def update_window(self, width: int, height: int, document_title=""):
+        self.resize(width, height)
+        if document_title != "":
+            post(Post.WINDOW_UPDATE_STATE, self.id, True, document_title)
+        else:
+            self.show()
```
FelipeDefensor, line 254: This code is shared by the three windows classes, right? Let's make a mixin class and inherit from that instead of duplicating the code in each class (see the `CursoMixIn` class in case you're not familiar with the pattern).

FelipeDefensor (changes requested):
Nice! Let's abstract the window functionality into a mixin (see coment) and I think we are good to go.

tilia/ui/timelines/pdf/timeline.py
azfoo: fixed now

FelipeDefensor:
The abstraction is perfect.

Found this bug: reopening the video window with a local media loaded results in a blank window.

https://github.com/TimeLineAnnotator/desktop/assets/87732202/7c64d169-d614-4cb5-940e-955619a44930

FelipeDefensor:
> The abstraction is perfect.
> 
> Found this bug: reopening the video window with a local media loaded results in a blank window.
> blank-reopened-video-window.mp4

Doesn't happen with YouTube player.

tilia/media/player/qtvideo.py
```diff
@@ -33,3 +24,10 @@ def _engine_load_media(self, media_path: str) -> bool:
     def _engine_exit(self):
         super()._engine_exit()
         self.widget.deleteLater()
+
+
+class QVideoWindow(ViewWindow, QVideoWidget):
+    def __init__(self):
+        super().__init__("TiLiA Player")
```
FelipeDefensor, line 31: image

Having the "TiLiA" before the "actual name" of the window looks very strange because it is redundant (we're already inside TiLiA) but I imagine the rationale is that this name will be also displayed by the OS window manager. I think we should pass two different strings to the constructor to solve this.

FelipeDefensor: Also, the menu string should make it explicit that this is the video player, not to be confused with the "player" "embedded" in the toolbar/slider timeline.

FelipeDefensor (changes requested):
See comments and the bug I found.

FelipeDefensor:
Oh, I missed the review request, sorry! I will review later today.

FelipeDefensor (approved):
The refactor looks good and everything works now. Feel free to squash and merge.

PR #156, opened 2024-07-12
No media

### LINKED ISSUE #153: Warn user that loading media failed when opening file
The current result is a silent fail. Notice that the media length is "loaded" nevertheless and that is by design, so the timelines can still be displayed even before loading a new media.

azfoo:
closes #153


FelipeDefensor (approved):
All good. Feel free to merge.

FelipeDefensor:
I forgot that I said that a waning alongside the disabled interface would be better. But seeing how it looks now, I don't think that is necessary anymore. The appearance of the player toolbar makes it  immediately clear that there is no media loaded.

PR #159, opened 2024-07-12
Fix double clicks on most timeline UI elements initiating drag

FelipeDefensor:


FelipeDefensor:
Fixes #158

azfoo:
I was trying to abstract this into the super class since there is a lot of "same" code - the app still works but the test fails. That could be a potential fix for the future.

FelipeDefensor:
> I was trying to abstract this into the super class since there is a lot of "same" code - the app still works but the test fails. That could be a potential fix for the future.

I was sure that I left a comment about that on the commit message body, but I can't find it. Yes, there is a lot of duplication going on. My first idea was to use a mixin since not all TimelineUIElement subclasses are draggable but abstracting into the super class might also work. I have created an issue draft in the project for this refactor.

azfoo (approved):
fixed!

PR #167, opened 2024-07-15
fix cli

azfoo:


azfoo:
probably fixed. requires further testing but is working now.

FelipeDefensor:
Good job! I have quickly tested it and everything seems to be working, apart from the "load-media" command. I will commit a fix later today.

FelipeDefensor (changes requested):
My last commits implement Player instances for the CLI to fix the 'load-media' command. In that case, do we still need to serve the media path from the App class? If not, I think it's a better idea to go back to serving from Player, as that has been tested manually previously.

Apart from that, it seems like we have got our CLI back. Nice! I have not tested it thorougly, but we can do that manually while preparing the analyses for the TISMIR article.

azfoo:
> My last commits implement Player instances for the CLI to fix the 'load-media' command. In that case, do we still need to serve the media path from the App class? If not, I think it's a better idea to go back to serving from Player, as that has been tested manually previously.
> 
Seems to work now with the previous method. I've updated and pushed things back up now.

FelipeDefensor (approved):
Looks good to me.

PR #174, opened 2024-07-16
Fix handling of quoted and spaced strings by the CLI

FelipeDefensor:


azfoo (approved):
Nice! works fine as far as I can tell.

PR #175, opened 2024-07-17
update csv import for beat

azfoo:
measure_number column:
- input number in column if measure number change is required. subsequent bars will count up from given number. will also cause ui to force display specified measure's number.

is_first_in_measure column:
- input `True` if given time is first beat of measure.

### DIFF
--- a/tests/parsers/csv/test_beats_from_csv.py
+++ b/tests/parsers/csv/test_beats_from_csv.py
@@ -39,3 +39,89 @@ def test_component_creation_fail_reason_gets_into_errors(beat_tl, tilia_state):
         )
 
     assert_in_errors("101", errors)
+
+
+def test_beats_from_csv_with_measure_number(beat_tlui):
+    os.chdir(Path(Path(__file__).absolute().parents[1]))
+
+    data = "time,measure_number\n5,1\n10,\n15,\n20,\n25,\n30,8"
+
+    tl = beat_tlui.timeline
+    with patch("builtins.open", mock_open(read_data=data)):
+        beats_from_csv(
+            tl,
+            Path(),
+        )
+
+    beats = sorted(tl.components)
+    assert beats[3].metric_position == (1, 4)
+    assert beats[4].metric_position == (8, 1)
+
+
+def test_beats_from_csv_with_is_first_in_measure(beat_tlui):
+    os.chdir(Path(Path(__file__).absolute().parents[1]))
+
+    data = "time,is_first_in_measure\n0,True\n5,\n10,\n15,\n20,\n25,True\n30,\n35,True"
+
+    tl = beat_tlui.timeline
+    with patch("builtins.open", mock_open(read_data=data)):
+        beats_from_csv(tl, Path())
+
+    beats = sorted(tl.components)
+    assert beats[4].metric_position == (1, 5)
+    assert beats[5].metric_position == (2, 1)
+    assert beats[6].metric_position == (2, 2)
+    assert beats[7].metric_position == (3, 1)
+
+
+def test_beats_from_csv_with_measure_number_and_is_first_in_csv(beat_tlui):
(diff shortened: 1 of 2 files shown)


tilia/parsers/csv/beat.py
```diff
@@ -18,40 +18,90 @@ def beats_from_csv(
...
     with TiliaCSVReader(path, file_kwargs, reader_kwargs) as reader:
-        params_to_indices = get_params_indices(["time"], next(reader))
+        params_to_indices = get_params_indices(params, next(reader))
 
         if "time" not in params_to_indices:
             display_column_not_found_error("time")
             return errors
 
+        if any(param in params_to_indices for param in optional_params):
+            beats_per_measure = (
```
FelipeDefensor, line 42: `beat_pattern` is a `list[int]` but it seems like you are treating it as an `int` in line 80. The idea of it being a list is to enable the user to set it to cycles like `4,3,3`.

FelipeDefensor (changes requested):
I like the solution. We really need some tests here as importing from CSV has proved to be error prone.

tilia/parsers/csv/beat.py
FelipeDefensor: Now I can see that it duplicates information, since the `is_first_in_measure` beat property already enables us to get that. That duplication leads to unnecessary complexity sometimes and maybe that is the case here. I would not oppose a refactor to eliminate the duplication in the future.

azfoo:
updated! now with tests and I also fixed the confusing list/0 variable.

tests/parsers/csv/test_beats_from_csv.py
```diff
@@ -39,3 +39,53 @@ def test_component_creation_fail_reason_gets_into_errors(beat_tl, tilia_state):
         )
 
     assert_in_errors("101", errors)
+
+
```
FelipeDefensor, line 43: Perhaps it would be nice to test the case where there is a `is_first_in_measure` column but there are no values in it. Also, what happens if the user inserts invalid values in that column?

FelipeDefensor (approved):
Nice solution. I'm fine with this the way it is, but make sure to consider my comments.

tests/parsers/csv/test_beats_from_csv.py
```diff
@@ -39,3 +39,53 @@ def test_component_creation_fail_reason_gets_into_errors(beat_tl, tilia_state):
...
+def test_beats_from_csv_with_measure_number(beat_tlui):
+    os.chdir(Path(Path(__file__).absolute().parents[1]))
+
+    data = "time,measure_number\n5,\n10,\n15,\n20,\n25,\n30,8"
```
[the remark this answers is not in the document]
azfoo: There isn't really a need for the line break because it's the last line. 30 is the time and 8 is the measure number.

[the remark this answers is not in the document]
FelipeDefensor: Wow, that was really silly of me 😅

tilia/parsers/csv/beat.py
```diff
@@ -18,39 +18,117 @@ def beats_from_csv(
...
+                    errors.append(
+                        "Time is not sorted in ascending order. Please sort before importing."
+                    )
+                    return errors
+                else:
+                    current_time = float(row[params_to_indices.get("time")])
+
+                if (
+                    params_to_indices.get("is_first_in_measure")
+                    and str(row[params_to_indices["is_first_in_measure"]]).lower()
```
FelipeDefensor, line 90: Do we need `str` here? I thought the csv reader parsed everything into strings.

FelipeDefensor (approved):
All the problems I had with importing the Huapstimme data seem to be fixed. I think you're done here 😃.

tilia/parsers/csv/beat.py
azfoo: good point.

PR #177, opened 2024-07-17
Fix csv docstrings

### LINKED ISSUE #176: Remove reference to CreateComponentErrors in csv importers docstrings
We do not raise that error when creating components anymore. Instead, we return a bool indicating success and a string providing the fail reason.

azfoo:
closes #176

### DIFF
--- a/tilia/parsers/csv/common.py
+++ b/tilia/parsers/csv/common.py
@@ -45,9 +45,18 @@ def _parse_attr_data(
                 else ""
             )
             if attr in required_attrs:
-                return False, [f"{value} is not a valid {attr.replace('_', ' ')}. " + appended_text], {}
+                return (
+                    False,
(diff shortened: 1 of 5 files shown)


tilia/parsers/csv/common.py
```diff
@@ -3,7 +3,7 @@
 from tilia.parsers.csv.base import AttributeData
 
 
-def _get_attrs_indices(params: list[str], headers: list[str]) -> [int]:
```
FelipeDefensor, line 6: There was a point in time when I thought this was the syntax for type hinting lists. We may found more of these lurking around the codebase 😅.

FelipeDefensor (approved):
I'm very lazy with docs, good to have someone around to keep me in check about it. Do this more, if you are up for it.

PR #179, opened 2024-07-19
fix typos

azfoo:


tests/ui/timelines/harmony/test_harmony_timeline_ui.py
```diff
@@ -6,8 +6,8 @@
 from tilia.requests import Get
 from tilia.ui.actions import TiliaAction
 
-FLAT_SIGN = "b"
-SHARP_SIGN = "#"
+FLAT_SIGN = "`b"
+SHARP_SIGN = "`#"
```
FelipeDefensor, line 10: I don't think we need the ` prefix with a single sign, but better safe than sorry.

FelipeDefensor (approved):
Thanks!

PR #180, opened 2024-07-19
Add relative times

### LINKED ISSUE #178: Enable customization of effective media boundaries

azfoo:
closes #178

### DIFF
--- a/tilia/errors.py
+++ b/tilia/errors.py
@@ -9,6 +9,10 @@ class Error(NamedTuple):
 
 
 FILE_SAVE_FAILED = Error("Save file", "Error when saving file.\n{}")
+METADATA_FIELD_INVALID = Error(
+    "Invalid metadata field",
+    "Metadata fields must be alphanumeric. The following fields were discarded:\n{}",
+)
 MEDIA_METADATA_IMPORT_JSON_FAILED = Error(
     "Import media metadata", "Error when parsing file {}:\n{}"
 )
--- a/tilia/errors.py
+++ b/tilia/errors.py
@@ -80,6 +80,10 @@ FILE_NOT_FOUND = Error(
 LOOP_DISJUNCT = Error("Loop Selection Error", "Selected Hierarchies are disjunct.")
 PLAYER_TOOLBAR_ERROR = Error("Updating Player Toolbar", "{}")
 YOUTUBE_PLAYER_ERROR = Error("Youtube Player Error", "{}")
+SEEK_OUTSIDE_CURRENT_PLAYBACK_TIME_FAILED = Error(
+    "Player seek failed",
+    "Selected time ({}) is outside of current time ({} - {}). Edit current start and end times in metadata to seek to this point.",
+)
 
 
 def display(error: Error, *args):
--- a/tilia/media/player/base.py
+++ b/tilia/media/player/base.py
@@ -5,6 +5,7 @@ import sys
 from abc import ABC, abstractmethod
 from enum import Enum, auto
 from pathlib import Path
+from dataclasses import dataclass
 
 from PyQt6.QtCore import QTimer
 
@@ -21,6 +22,7 @@ from tilia.requests import (
     get,
     stop_serving_all,
 )
+from tilia.ui.format import format_media_time
 
 
 class MediaTimeChangeReason(Enum):
@@ -39,10 +41,9 @@ class Player(ABC):
 
         self._setup_requests()
         self.is_media_loaded = False
-        self.duration = 0.0
-        self.playback_start = 0.0
-        self.playback_end = 0.0
         self.current_time = 0.0
+        self.absolute_time = TimeRange()
+        self.playback_time = TimeRange()
         self.media_path = ""
         self.is_playing = False
         self.is_looping = False
@@ -70,12 +71,16 @@ class Player(ABC):
             (Post.PLAYER_REQUEST_TO_LOAD_MEDIA, self.load_media),
             (Post.PLAYER_EXPORT_AUDIO, self.on_export_audio),
             (Post.PLAYER_CURRENT_LOOP_CHANGED, self.on_loop_changed),
+            (Post.MEDIA_TIMES_PLAYBACK_UPDATED, self.setup_playback_start_and_end),
+            (Post.FILE_MEDIA_DURATION_CHANGED, self.on_file_media_duration_changed),
         }
 
         SERVES = {
             (Get.MEDIA_CURRENT_TIME, lambda: self.current_time),
             (Get.MEDIA_PATH, lambda: self.media_path),
             (Get.MEDIA_TYPE, lambda: self.MEDIA_TYPE),
+            (Get.MEDIA_TIMES_ABSOLUTE, lambda: self.absolute_time),
+            (Get.MEDIA_TIMES_PLAYBACK, lambda: self.playback_time),
         }
 
         for post, callback in LISTENS:
@@ -86,11 +91,9 @@ class Player(ABC):
 
     @property
     def playback_length(self):
-        return self.playback_end - self.playback_start
+        return self.playback_time.duration
 
-    def load_media(
-        self, path: str | Path, start: float = 0.0, end: float = 0.0
-    ) -> bool:
+    def load_media(self, path: str | Path) -> bool:
         if self.is_playing:
             self.stop()
 
@@ -98,40 +101,51 @@ class Player(ABC):
         if not success:
             tilia.errors.display(tilia.errors.MEDIA_LOAD_FAILED, path)
             return False
-        self.on_media_load_done(path, start, end)
+        self.on_media_load_done(path)
         return True
 
-    def on_media_load_done(self, path, start, end):
+    def on_media_load_done(self, path):
         self.media_path = str(path)
-        self.playback_start = start
-
-        post(
-            Post.PLAYER_URL_CHANGED,
-            self.media_path,
-        )
-
-        post(Post.PLAYER_CURRENT_TIME_CHANGED, 0.0, MediaTimeChangeReason.LOAD)
+        post(Post.PLAYER_URL_CHANGED, self.media_path)
 
         self.is_media_loaded = True
+        self._engine_seek(self.playback_time.start)
+        self.current_time = self.playback_time.start
+        post(
+            Post.PLAYER_CURRENT_TIME_CHANGED,
+            self.current_time,
+            MediaTimeChangeReason.LOAD,
+        )
 
-    def on_media_duration_available(self, duration):
-        self.playback_end = self.duration = duration
-        post(Post.PLAYER_DURATION_AVAILABLE, duration)
+    def on_file_media_duration_changed(self, duration: float, start: float, end: float):
+        self.absolute_time.end = duration
+        if start > 0.0 or end < duration:
+            self.setup_playback_start_and_end(start, end)
+        else:
+            self.playback_time.end = duration
+            post(Post.PLAYER_DURATION_AVAILABLE, self.playback_time.duration)
+
+    def on_media_duration_available(self, duration: float):
+        self.absolute_time.end = self.playback_time.end = duration
+        post(Post.PLAYER_DURATION_AVAILABLE, self.playback_time.duration)
 
     def setup_playback_start_and_end(self, start, end):
-        self.playback_start = start
-        self.playback_end = end or self.duration
-
-        start or self._engine_seek(start)
+        self.playback_time.start = start
+        self.playback_time.end = end
+        post(Post.PLAYER_DURATION_AVAILABLE, self.playback_time.duration)
+        self.on_seek(start)
+        self.is_looping = False
+        post(Post.PLAYER_CANCEL_LOOP)
 
     def unload_media(self):
         self._engine_unload_media()
         self.is_media_loaded = False
-        self.duration = 0.0
         self.current_time = 0.0
         self.media_path = ""
         self.is_playing = False
         self.is_looping = False
+        self.absolute_time = TimeRange()
+        self.playback_time = TimeRange()
         post(Post.PLAYER_CANCEL_LOOP)
         post(Post.PLAYER_MEDIA_UNLOADED)
 
@@ -153,7 +167,7 @@ class Player(ABC):
     def stop(self):
         """Stops music playback and resets slider position"""
         post(Post.PLAYER_STOPPING)
-        if not self.is_playing and self.current_time == 0.0:
+        if not self.is_playing and self.current_time == self.playback_time.start:
             return
 
         self._engine_stop()
@@ -164,8 +178,8 @@ class Player(ABC):
             self.is_looping = False
             post(Post.PLAYER_CANCEL_LOOP)
 
-        self._engine_seek(self.playback_start)
-        self.current_time = self.playback_start
+        self._engine_seek(self.playback_time.start)
+        self.current_time = self.playback_time.start
 
         post(Post.PLAYER_STOPPED)
         post(
@@ -176,8 +190,8 @@ class Player(ABC):
 
     def on_loop_changed(self, start_time, end_time):
         self.is_looping = start_time != end_time
-        if end_time == get(Get.MEDIA_DURATION):
-            if start_time != 0:
+        if end_time == self.playback_time.end:
+            if start_time != self.playback_time.start:
                 end_time -= self.E
             else:
                 self._engine_loop(True)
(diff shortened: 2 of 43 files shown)


tilia/errors.py
```diff
@@ -76,6 +76,10 @@ class Error(NamedTuple):
 LOOP_DISJUNCT = Error("Loop Selection Error", "Selected Hierarchies are disjunct.")
 PLAYER_TOOLBAR_ERROR = Error("Updating Player Toolbar", "{}")
 YOUTUBE_PLAYER_ERROR = Error("Youtube Player Error", "{}")
+SEEK_OUTSIDE_CURRENT_RELATIVE_TIME_FAILED = Error(
+    "Player seek failed",
+    "Seeked time ({}) is outside of current time ({} - {}). Edit current start and end times in metadata to seek to this point.",
```
FelipeDefensor, line 85: I think the more user-friendly term here is "selected time". The second use of seek is fine, by me, as what came before makes it clear to what we are referring, evne if the user doesn't know the mor etechnical term for what he is doing.

tilia/media/player/base.py
```diff
@@ -352,3 +365,31 @@ def _engine_loop(self, is_looping: bool) -> None:
 
     def __repr__(self):
         return f"{type(self)}-{id(self)}"
+
+
+class Time:
```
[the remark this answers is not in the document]
FelipeDefensor: Also, this would be a good place for using dataclasses, in case you have heard of them. No need to change it, though.

tilia/ui/coords.py
```diff
@@ -1,11 +1,18 @@
...
-def get_x_by_time(time: float) -> int:
+def get_x_by_time(time: float) -> float | None:
     try:
-        return ((time / get(Get.MEDIA_DURATION)) * get(Get.PLAYBACK_AREA_WIDTH)) + get(
-            Get.LEFT_MARGIN_X
-        )
+        relative_time = get(Get.MEDIA_TIMES_RELATIVE)
+        if (
+            time < relative_time.start
+            or time > get(Get.MEDIA_DURATION) + relative_time.start
```
FelipeDefensor, line 9: Again, chained comparisons would help to make this a bit more readable.

tilia/ui/timelines/harmony/elements/harmony.py
```diff
@@ -306,7 +306,11 @@ def get_point(self, x: float, y: float):
         return QPointF(x - self.boundingRect().width() / 2, y)
 
     def set_position(self, x, y):
-        self.setPos(self.get_point(x, y))
+        if x == None:
```
FelipeDefensor, line 309: Minor thing: The recommended way to compare with `None` is through the `is` operator, as in `if x is None`. We should check for identity, and not equality, as `None` is a singleton in Python (https://stackoverflow.com/a/25317251/15862653).

tilia/ui/timelines/hierarchy/element.py
```diff
@@ -311,11 +311,14 @@ def update_loop_icon_position(self):
         )
 
     def update_label_position(self):
-        self.label.set_position(
-            (self.start_x + self.end_x) / 2,
-            self.timeline_ui.get_data("height"),
-            self.get_data("level"),
-        )
+        if self.start_x != None and self.end_x != None:
```
FelipeDefensor, line 314: As there are multiple comparisons with None, I recommend changing them all according to the comment above.

tilia/ui/windows/metadata.py
```diff
@@ -82,6 +88,44 @@ def populate_with_read_only_fields(self):
             value_label.setWordWrap(True)
             self.form_layout.addRow(QLabel(name.capitalize()), value_label)
 
+    def populate_with_media_times(self):
```
FelipeDefensor, line 91: Do you think it would be nice to also display the absolute media length? I think it might m be useful if the user wants to reset the effective end to the absolute length.

FelipeDefensor (changes requested):
As always, overall this looks great. I have yet to test things more thoroughly, but everything seems to be working.I made minor code style comments, mostly.

There is only one relevant design issue to discuss. I'm not sure about hiding the components if changing the effective boundaries gets them out of bounds. I can't see a use case where the user would find out-of-bounds, invisible components useful, and I imagine that making them invisible might make the user forget to edit them, and potentially leave them outdated. Also, it adds complexity on our side, having to deal with those `time is None` cases. What do you think about prompting the user and deleting the out-of-bounds elements, instead?

tilia/media/player/base.py
```diff
@@ -199,6 +203,15 @@ def on_seek(self, time: float, if_paused: bool = False) -> None:
             return
 
         if self.is_media_loaded:
+            if time < self.relative_time.start or time > self.relative_time.end:
```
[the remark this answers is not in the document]
azfoo: sorry, I forgot how to python 😅

FelipeDefensor:
Actually, you can't used chained comparisons there as the operator is an `or`. Apparently you still remember you Python :))

tilia/ui/windows/metadata.py
azfoo: The main reason why I went with hiding is that AudiowaveTimeline reads the entire file and displays components for everything - that's something that can be changed in the timeline I guess. The other use case I was thinking of was for showing specific parts of the file and hiding the rest. (I'm hoping YouTube will update their API soon to include chapters and all the fun things you can do with that).

But happy to not have the None - perhaps a tuple to show whether it is out-of-bounds?

azfoo: sound. will do.

azfoo:
The main reason for hiding is in the AudiowaveTimeline because that displays components for the entire file - easily changeable, so we could go without the None. 
_scale_or_crop clears up all the other timelines anyway so I think we could technically do without the checks in the other timelines. 
Alternatively, I was thinking get_x_by_time could return a tuple (value, whether it is out of bounds) so we don't have mixed datatypes.

FelipeDefensor:
> The main reason for hiding is in the AudiowaveTimeline because that displays components for the entire file - easily changeable, so we could go without the None. _scale_or_crop clears up all the other timelines anyway so I think we could technically do without the checks in the other timelines. Alternatively, I was thinking get_x_by_time could return a tuple (value, whether it is out of bounds) so we don't have mixed datatypes.

Ok. So you are saying that we could use the `_scale_or_crop` to delete the out-of-bounds elements? If so, great. I like not having mixed types and I am a fan of returning tuples with bool flags to indicate success and the like, you will see those all around the codebase, so I think that is the way to go if we are going with the inbound check. I still prefer avoiding the complexity in this case if we can, though.

FelipeDefensor:
Hi, Anne. I have found some bugs.

---
Initial hierarchy is not visible when creating hierarhcy timeline after setting effective boundary != 0.

Steps to reproduce:
- New file
- Load media
- Set effective start != 0
- Create hierarchy timeline
---
Effective times are not being loaded from saved file.

Steps to reproduce:
- New file
- Load media
- Set effective start != 0
- Set effective end != media length
- Save file
- Open same file

Result: Effective start is set to 0 and effective end is set to media length.
---
Creating new file after setting effective start has wrong time displayed

Steps to reproduce:
- New file
- Load media
- Set effective start to 10
- Create new file with File > New file

Result: Displayed time is 59:50.0/59:50.0

azfoo:
not quite fixed. I'm slightly stuck on how to serve get.media_metadata in the failing tests. could you help?

FelipeDefensor:
> not quite fixed. I'm slightly stuck on how to serve get.media_metadata in the failing tests. could you help?

Sure! I will take at a look at this shortly.

FelipeDefensor:
> > not quite fixed. I'm slightly stuck on how to serve get.media_metadata in the failing tests. could you help?
> 
> Sure! I will take at a look at this shortly.

The problem is that due to how I implemented the test teardown, at the moment `tilia_state` tries to reset the duration to the default 100, `file_manager` has already been torn down, and, with that, has stop serving `Get.MEDIA_METADATA`. However, resetting the duration to 100 involves calling `App.set_media_duration` which posts `Post.FILE_MEDIA_DURATION_CHANGED` which triggers `Player.on_media_duration_available`. That last method tries to get `Get.MEDIA_METADATA`, but that's not available because, as said above, the file manager has already been torn down.

I'm not sure how to fix this yet, but couldn't we just store the relative times in the player itself and evade the need to ask the file manager for them? Actually, don't we already do that in `Player.relative_time`?

azfoo:
That should fix everything

tilia/timelines/collection/collection.py
```diff
@@ -307,55 +313,70 @@ def get_timeline_ids(self):
     def has_timeline_of_kind(self, kind: TlKind):
         return any([tl.KIND == kind for tl in self])
 
-    def _scale_or_crop_timelines(self, new_duration, prev_duration):
+    def _scale_or_crop_timelines(self):
```
FelipeDefensor, line 316: I think this function has too many responsabilities. Sorry for suggesting that we reuse it in this context. Asking if a user wants to rescale the timelines doesn't really make sense when changing the effective boundaries. As the absolute size of the media hasn't changed, I can't see any reason for wanting to scale the components, as they were surely made accodirng to the appropriate (and only) scale in the first place.

Asking if the user wants to crop makes sense though, as component that were in bounds might get out of bounds if we change the effective range.

The more durable solution would probably be to split scaling and cropping logic into separate functions and call both when loading a new media but just the crop part when changing the effective boundaries.

tilia/timelines/collection/collection.py
```diff
@@ -307,55 +313,70 @@ def get_timeline_ids(self):
...
 
-        self.cached_media_duration = new_duration
+        self.cached_media_start = relative_time.start
+        self.cached_media_end = relative_time.end
 
     def serve_ordinal_for_new_timeline(self):
         return len(self._timelines) + 1
 
-    def scale_timeline_components(self, factor: float) -> None:
+    def scale_timeline_components(self, offset: float, factor: float) -> None:
```
FelipeDefensor, line 373: I think I get what the offset does but I can't see a use case. I'm doubly sorry for not commenting on the scaling before, as that would have maybe saved you the trouble of adding this.

I may be mistaken about removing the scale option when changing effective boundaries, though. Please do say if you think this is a useful feature. I would rather be wrong in this case, actually.

FelipeDefensor (changes requested):
I think the bugs have all been solved, but this has a lot of moving parts. See my comments.

I also want to discuss the "relative times" name. I'm not sure what they would be relative too. "Absolute time" does, on the other hand, sound like a good name for the "file" boundaries. As alternatives to "relative start/end", we have "playback start/end", "annotation start/end" and "timelines start/end". We could also go with setting "start offset" and "end offset" instead, but that would require adaoting some of the code. What are your thoughts on this?

azfoo:
> I also want to discuss the "relative times" name. I'm not sure what they would be relative too. "Absolute time" does, on the other hand, sound like a good name for the "file" boundaries. As alternatives to "relative start/end", we have "playback start/end", "annotation start/end" and "timelines start/end". We could also go with setting "start offset" and "end offset" instead, but that would require adaoting some of the code. What are your thoughts on this?
>
I like playback start/end best. Naming it offset would mean more maths (which isn't too bad) but it is a strange way of thinking about time - also that it would involve checks to make sure the boundaries don't overlap.

tilia/timelines/collection/collection.py
azfoo: I think scale could potentially be useful in the context of the user trying to compare two different media sources and seeing how they line up with each other - eg. if the rate of change in tempo is consistent, not the actual tempo - and creating comparative charts from there. 
Another strange use could be if the user found a more accurate remaster of an old warbly recording that is less warbly and re-recorded at the right rpm - in that case, the time ratio between elements will remain the same.

azfoo: offset is the c of y = mx + c - it corrects the start time in case it is non 0.

FelipeDefensor: > I think scale could potentially be useful in the context of the user trying to compare two different media sources and seeing how they line up with each other - eg. if the rate of change in tempo is consistent, not the actual tempo - and creating comparative charts from there. Another strange use could be if the user found a more accurate remaster of an old warbly recording that is less warbly and re-recorded at the right rpm - in that case, the time ratio between elements will remain the same.

I agree, but in both cases they would be loading a new media and not changing the playback start/end of the current one, right? I'm not suggesting that we remove scaling, but that we only prompt for that when loading new media.

azfoo: I see what you mean - but if the new file were not previously cropped exactly to where things should start and end, there wouldn't be a way for the app to crop it internally and scale. (Am I still misunderstanding something?😅)

FelipeDefensor:
> > I also want to discuss the "relative times" name. I'm not sure what they would be relative too. "Absolute time" does, on the other hand, sound like a good name for the "file" boundaries. As alternatives to "relative start/end", we have "playback start/end", "annotation start/end" and "timelines start/end". We could also go with setting "start offset" and "end offset" instead, but that would require adaoting some of the code. What are your thoughts on this?
> 
> I like playback start/end best. Naming it offset would mean more maths (which isn't too bad) but it is a strange way of thinking about time - also that it would involve checks to make sure the boundaries don't overlap.

Ok, I like playback start/end too. Let's go with that.

tilia/timelines/collection/collection.py
FelipeDefensor: Oh, you mean in the case we are loading a file to a analysis that already has playback start != 0 and/or playback end != media length?

If that's the case, I have to say I hadn't thought about it. How are you planning on dealing with that?

If that's not the case, then I'm being a little slow to understand what you're saying and will have to ask you to explain it again 😅.

FelipeDefensor: Oh, now I get it! This does seem useful, so let's keep the scaling option. I think we should improve the confirmation dialog, then, I believe the current one may come as a surprise to a user that is only looking to change the playback start/end. How about something like this?

image

azfoo: Are you implying the current separate scale/crop dialogues get squished into one? That seems much more reasonable than asking twice, although the dialogues for a longer/shorter new time will have to be different.

I had plans for building the RangeSlider (implemented in QML but not ported over to PyQt) - which will be nice for sliding both start and end points in setting times in metadata and here. But that will be a plan for much later because it looks to be potentially complicated.

FelipeDefensor: I forgot they were separate, so yes, that's what I suggest. Which one of the two I sent gets displayed depends on if the new time is longer or shorter, like you said.

A RangeSlider would definitely help, but we can certainly wait.

tilia/ui/timelines/audiowave/element.py
```diff
@@ -127,17 +122,22 @@ def on_deselect(self) -> None:
...
-        self.width = width / 100
-        self.selected = False
-        self.update_pen_style()
+        self.set_position(start_x, width, amplitude, height)
 
     def set_position(self, start_x, width, amplitude, height):
         self.setLine(self.get_line(start_x, amplitude, height))
-        self.width = width / 100
+        self.width = width
+        self.selected = False
```
FelipeDefensor, line 140: This makes so that every time we redraw (e.g. when zooming) the component gets deselected. It should go back to the initializer. The rest is fine.

FelipeDefensor (changes requested):
## New dialog
I really like the dialog! It's much clearer than the previous one. 

## Bugs
Given the complexity that this feature adds, it's no surprise that I found a few bugs:

- Hierarchy positions are not being updated after increasing duration and choosing the "fill with empty space" option
- Decreasing duration and compressing timelines twice in a row break hierarchy timelines
- Decreasing end time and cropping followed by increasing start and cropping break hierarchy timelines (see video below)




https://github.com/user-attachments/assets/6a9ea434-816a-4155-acba-9821f5425127


## PR process
The size of the code that GitHub prompts me to review has gotten out of hand on this PR. If I understand correctly, you have done something that rewrote the git's history and had to do a `git push --force` after the last review. Doing that makes so that all the PR's commits will be marked for review on the next review request. I think the ideal thing would be to defer anything that rewrites git's history to after the PR has been approved and right before merging it. Can we do that?

FelipeDefensor:
I changed the PR destination to the v0.5 branch. As we didn't get this done before starting the manual testing, I think it's safer to postpone the release. We can still use relative times in the proofs-of-concept for the article by setting them directly on the `.tla`. I will have to check if it will be possible to open them (after downloading from the website) on the v0.4.1 desktop app, though.

azfoo:
> > Regarding PR process, most of the rewrites happen because of rebasing to upstream. Would you prefer if I didn't rebase until everything is finished?
> 
> Let's try that and see if it makes things easier, yes.

My bad, I did it again... It is only the last two commits that are actually updates.

FelipeDefensor:
> > > Regarding PR process, most of the rewrites happen because of rebasing to upstream. Would you prefer if I didn't rebase until everything is finished?
> > 
> > 
> > Let's try that and see if it makes things easier, yes.
> 
> My bad, I did it again... It is only the last two commits that are actually updates.

No problem, thanks for the warning.

Just to make sure: you're still working on those bugs so I still shouldn't review, right?

azfoo:
sorry for the delay - bugs have now been resolved

PR #183, opened 2024-07-22
feat: add save metadata buttons

### LINKED ISSUE #182: Missing a save button in the metadata window
Currently, the only way to save is to confirm the dialog when trying to close, which is counterintuitive. We could without the save button if changes were autosaved, but hen we would need a discard button.

azfoo:
closes #182

### DIFF
--- a/tilia/ui/windows/metadata.py
+++ b/tilia/ui/windows/metadata.py
@@ -6,7 +6,10 @@ from PyQt6.QtWidgets import (
     QLabel,
     QLineEdit,
     QPushButton,
-    QFrame
+    QFrame,
+    QVBoxLayout,
+    QWidget,
+    QDialogButtonBox,
 )
 
 from tilia.settings import settings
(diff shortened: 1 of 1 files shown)


FelipeDefensor (approved):
All good!

FelipeDefensor:
So, now that there's no exit confirmation dialog and we need to apply the changes, the initial focus on the cancel button is an issue. If we type into any field and press Enter hoping that it would apply the changes or take us to the next field, it actually closes the window immediately without saving.

I think we need a confirmation for discarding changes. Additionally, it might be a good idea to remove the focus from the cancel button.

azfoo:
How about putting the confirm close back onto the close event and leaving the focus on cancel? (I think the focus on cancel is set by the os by default). That way even if you accidentally press cancel, you'll still have to go through another dialog before you actually close the window.

FelipeDefensor:
> How about putting the confirm close back onto the close event and leaving the focus on cancel? (I think the focus on cancel is set by the os by default). That way even if you accidentally press cancel, you'll still have to go through another dialog before you actually close the window.

Yes, that works!

PR #185, opened 2024-07-26
feat: add confirm dialog before close

### LINKED ISSUE #182: Missing a save button in the metadata window

azfoo:
#182 
Reimplemented confirm dialog before close


FelipeDefensor (approved):
👍🏼

FelipeDefensor:
Oh, I forgot I was in you fork when doing that last commit, sorry. We can merge that too.

PR #186, opened 2024-07-29
feat: script generator

azfoo:
~~mostly~~ working. 
~~save is broken - path.parent.exists() fails even though path does exist.~~
~~having spaces in directory names is also another cause of failure.~~

### DIFF
--- a/tests/ui/cli/test_generate_scripts.py
+++ b/tests/ui/cli/test_generate_scripts.py
@@ -1,0 +1,100 @@
+from tilia.ui.cli.generate_scripts import (
+    _get_args_from_filename,
+    _get_script_for_folder,
+)
+import pytest
+
+
+@pytest.mark.parametrize(
+    "filename, kind, expected_args, expected_is_by_time, expected_name",
+    [
+        (
+            "timeline_beat_count_beat-pattern_4_height_20",
+            "beat",
+            ["--beat-pattern 4", "--height 20"],
+            True,
+            "timeline_count",
+        ),
+        (
+            "timeline_hrc_hierarchy__height_20_by-time",
+            "hrc",
+            ["--height 20"],
+            True,
+            "timeline_hierarchy_",
+        ),
+        ("timeline_mrk_name_by-measure", "mrk", [], False, "timeline_name"),
+    ],
+)
+def test_get_args_from_filename(
+    filename, kind, expected_args, expected_is_by_time, expected_name
+):
+    args, is_by_time, name = _get_args_from_filename(filename, kind)
+    for arg in expected_args:
+        assert arg in args
+    assert is_by_time == expected_is_by_time
+    assert name == expected_name
+
+
+def test_get_args_from_filename_with_time_and_measure_fails():
+    with pytest.raises(
+        ValueError, match=r"File name contains both by-time and by-measure."
+    ):
+        _get_args_from_filename("marker_by-time_by-measure", "marker")
+
+
+def test_get_script_for_folder(tmpdir):
+    duration = 250
+    with open(tmpdir.join("set_media_length.txt"), "w") as f:
+        f.write(str(duration))
+
+    filenames = [
+        "some_media_file.mp3",
+        "beat_tlName.csv",
+        "metadata.json",
+        "set_media_length.txt",
+    ]
+    output_script_path = _get_script_for_folder(tmpdir.strpath, filenames)
+
+    assert output_script_path == tmpdir.join("script.txt")
+
+    output_script = tmpdir.join("script.txt").read()
+
+    assert output_script_path == tmpdir.join("script.txt")
+    assert f'load-media "{tmpdir.join("some_media_file.mp3")}"' in output_script
+    assert f"metadata set-media-length {duration}" in output_script
+    assert f'metadata import "{tmpdir.join("metadata.json")}"' in output_script
+    assert f'timelines add beat --name "tlName"' in output_script
+    assert (
+        f'timelines import csv beat --target-name "tlName" --file "{tmpdir.join("beat_tlName.csv")}"'
+        in output_script
+    )
+    assert f'save "{tmpdir}' in output_script and '.tla"' in output_script
+
+
+def test_get_script_for_folder_returns_none_on_no_media(tmpdir):
+    filenames = ["beat_tlName.csv", "metadata.json"]
+    assert _get_script_for_folder(tmpdir.strpath, filenames) == None
+
+
+def test_get_script_for_folder_required_beat_file(tmpdir):
+    filenames = [
+        "some_media.mp4",
+        "beat_tlName.csv",
+        "metadata.json",
+        "marker_tlName2_by-time.csv",
+    ]
+    assert _get_script_for_folder(tmpdir.strpath, filenames) != None
+    assert "by-time" in tmpdir.join("script.txt").read()
+
+    filenames = [
+        "some_media.mp4",
+        "beat_tlName.csv",
+        "metadata.json",
+        "marker_tlName2_by-measure.csv",
+    ]
+    assert _get_script_for_folder(tmpdir.strpath, filenames) != None
+    assert "by-measure" in tmpdir.join("script.txt").read()
+    assert "--reference-tl-name" in tmpdir.join("script.txt").read()
+
+    filenames = ["some_media.mp4", "metadata.json", "marker_tlName2_by-measure.csv"]
+    assert _get_script_for_folder(tmpdir.strpath, filenames) == None
--- a/tilia/media/constants.py
+++ b/tilia/media/constants.py
@@ -6,3 +6,11 @@ CONVERTIBLE_AUDIO_FORMATS = [
     "flac",
 ]  # ffmpeg can probably convert more formats
 SUPPORTED_VIDEO_FORMATS = ["mp4", "mkv", "m4a"]
+ALL_SUPPORTED_MEDIA_FORMATS = {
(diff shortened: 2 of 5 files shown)


import_timelines.py
```diff
@@ -0,0 +1,172 @@
...
+            return (
+                f"timelines add {kind} --name {name}{(" " + args) if args else " "}",
+                f'timelines import csv {kind}{time_or_measure}--target-name {name} --file "{os.path.join(folder_name, filename)}"',
+                has_beat,
+                requires_beat,
+                reference_beat,
+            )
+
+
+def run():
```
FelipeDefensor, line 88: The six indentations levels make this hard to read. Let's refactor. I will leave some ideas in specific parts below.

import_timelines.py
```diff
@@ -0,0 +1,172 @@
...
+        List of paths to individual scripts (str): The absolute path to each script.
+    """
+    directory = _get_starting_directory()
+    saved_scripts = []
+    for folder_name, sub_folders, filenames in os.walk(directory):
+        media_data = _get_media_files(filenames)
+        if not (media_data or "set_media_length.txt" in filenames):
+            print(f"No suitable media found in {folder_name}.")
+        elif len(media_data) > 1:
+            print(f"Multiple media files found in {folder_name}: {media_data}")
```
FelipeDefensor, line 109: We could turn these into guard clauses to avoid the extra indentation level and make clear that these are "error" cases.

import_timelines.py
```diff
@@ -0,0 +1,172 @@
...
+                if media_data:
+                    f.write(
+                        f'load-media "{os.path.join(folder_name, media_data[0])}"\n'
+                    )
+                    filenames.remove(media_data[0])
+                if "set_media_length.txt" in filenames:
+                    f.write(
+                        f'metadata set-media-length {open(os.path.join(folder_name, "set_media_length.txt"), "r").read()}\n'
+                    )
+                    filenames.remove("set_media_length.txt")
```
FelipeDefensor, line 121: Ideally, there should be no logic in the `open` context manager, just the writing itself. You can compute the strings before entering it.

import_timelines.py
```diff
@@ -0,0 +1,172 @@
...
+                            requires_beat |= check_requires_beat
+                            if reference_beat == "" and check_reference_beat:
+                                reference_beat = check_reference_beat
+                            elif reference_beat != "" and check_reference_beat:
+                                print(
+                                    f"Multiple beat timelines found. Using {reference_beat} as reference instead of {check_reference_beat}"
+                                )
+
+                        except ReferenceError as e:
+                            print(e)
```
FelipeDefensor, line 151: This is a good candidate for extraction into a separate function.

import_timelines.py
```diff
@@ -0,0 +1,172 @@
...
+                            (
+                                add,
+                                imp,
+                                check_has_beat,
+                                check_requires_beat,
+                                check_reference_beat,
+                            ) = _get_timeline_args(file, folder_name)
+                            f.write(f"{add}\n")
+                            timelines_to_import.append(imp)
+                            has_beat |= check_has_beat
```
FelipeDefensor, line 141: Thanks for letting me know about this operator 😃

import_timelines.py
```diff
@@ -0,0 +1,172 @@
...
+                            requires_beat |= check_requires_beat
+                            if reference_beat == "" and check_reference_beat:
+                                reference_beat = check_reference_beat
+                            elif reference_beat != "" and check_reference_beat:
+                                print(
+                                    f"Multiple beat timelines found. Using {reference_beat} as reference instead of {check_reference_beat}"
+                                )
+
+                        except ReferenceError as e:
```
FelipeDefensor, line 150: Are you sure this is not supposed to a `KeyError` or `ValueError`? `ReferenceError` only happens when using a `WeakRefDict` AFAIK.

import_timelines.py
```diff
@@ -0,0 +1,172 @@
...
+                            print(e)
+
+                if requires_beat and not has_beat:
+                    print(f"{folder_name} requires beat but no beat csv found.")
```
FelipeDefensor, line 154: Another guard clause candidate.

FelipeDefensor (commented):
Nice!

This is a standalone script, right? If so, I think we should host it on a `tilia-utils` repo. An alternative would be to integrate this into the CLI.

Design-wise, I would really like to see `run` broken up in three or four functions for readability. See my comments.

If the idea is to mention this on the article, I think you can go ahead and do that while we polish this and figure out where it should go.

import_timelines.py
azfoo: Agreed. Was trying to work out where to take things out...

import_timelines.py
azfoo: Realistically, we could name it whatever we want as long as it makes sense to the user. I was having a hard time deciding what kind of error it was. Do you have a preference between Key and Value?

azfoo:
> This is a standalone script, right? If so, I think we should host it on a `tilia-utils` repo. An alternative would be to integrate this into the CLI.
> 
I wasn't sure where this should live (certainly not where it currently lives). If you think it belongs on a different repo, you'll need to move it because I don't have access to making new repos. 
Putting it on the CLI, on the other hand, would make it easier to access - where if the directory were you thinking?

There's another WIP on automating script execution once they are created - would that be helpful?

> If the idea is to mention this on the article, I think you can go ahead and do that while we polish this and figure out where it should go.
>
I honestly don't know if it should go in, just thought it would be a useful tool to have instead of having to write out multiple scripts.

import_timelines.py
FelipeDefensor: Oh, this is an error we raise, ok. I would say we go with ValueError.

FelipeDefensor:
> I wasn't sure where this should live (certainly not where it currently lives). If you think it belongs on a different repo, you'll need to move it because I don't have access to making new repos. Putting it on the CLI, on the other hand, would make it easier to access - where if the directory were you thinking?

Let's get that permission granted in any case, then. I was thinking we could have a command like `script generate --recurse` if we're going with the CLI, so that would mean refactoring `script.py` into a separate folder and having a `generate.py` inside it. The downside to integrating this into the CLI is that we're going to have to commit to testing and maintaining this (complex) functionality and some adptation to the CLI code would be needed. If you feel like tackling this now, go for it. Otherwise, we can have this as a standalone and integrate it later, when we need it.

> There's another WIP on automating script execution once they are created - would that be helpful?

Yes, I do think it will be helpful once we get into converting large corpora (soon, I hope).
 
> I honestly don't know if it should go in, just thought it would be a useful tool to have instead of having to write out multiple scripts.

Fair enough. I think there's a point where a mention would fit in nicely. I'll let you know.

import_timelines.py
```diff
@@ -0,0 +1,172 @@
...
+        if arg in split_name:
+            " ".join([args, f"--{arg} {split_name.pop(split_name.index(arg) + 1)}"])
+            split_name.remove(arg)
+
+    if "by-measure" in split_name:
+        is_by_time = False
+        split_name.remove("by-measure")
+
+    if "by-time" in split_name:
+        if "is_by_time" in locals():
```
FelipeDefensor, line 46: This is certainly valid Python, and it's nice to know about `locals`, but it's not idiomatic to rely on such introspection (not sure if that's the right term here) when there are reasonable alternatives. More specifically, Python programs don't usually check if variables are defined before using them. They either are defined and everyone is happy or they're not and we get a `ValueError`. You could have an additional `is_by_measure` bool and check if that's true at this point, instead.

azfoo:
> > I wasn't sure where this should live (certainly not where it currently lives). If you think it belongs on a different repo, you'll need to move it because I don't have access to making new repos. Putting it on the CLI, on the other hand, would make it easier to access - where if the directory were you thinking?
> 
> Let's get that permission granted in any case, then. I was thinking we could have a command like `script generate --recurse` if we're going with the CLI, so that would mean refactoring `script.py` into a separate folder and having a `generate.py` inside it. The downside to integrating this into the CLI is that we're going to have to commit to testing and maintaining this (complex) functionality and some adptation to the CLI code would be needed. If you feel like tackling this now, go for it. Otherwise, we can have this as a standalone and integrate it later, when we need it.
> 
The only dependence on TiLiA at the moment is the constants file to get acceptable media types. Bundling this into the app would make overall access easier to manage. Linking it to scripts would make managing the looping easier as well. 
If "tackling this now" refers to the more flexible due date that is not tomorrow, it probably is doable.

import_timelines.py
```diff
@@ -0,0 +1,172 @@
...
+        if os.path.exists(directory) and os.path.isdir(directory):
+            return directory
+        else:
+            print("Please ensure input is an existing directory.")
+
+
+def _get_media_files(filenames: list[str]) -> list:
+    supported_formats = [
+        file_format
+        for media_type, values in vars(constants).items()
```
FelipeDefensor, line 18: Other unrelated constants might get added to that file, so it's better to import the formats explicitly here. If you want a more concise import, you could create an `ALL_SUPPORTED_MEDIA_FORMATS` constant in `tilia.media.constants` concatening the existing ones.

FelipeDefensor: Maybe this also qualifies as a kind of instrospection that's better to avoid (see comment below).

FelipeDefensor (changes requested):
The whole thing is much more readable, thanks! Couple issues with using `vars` and `locals`, only.

FelipeDefensor (approved):
Looks good to me. Nice job on the tests!

PR #187, opened 2024-08-06
Fix zoom

azfoo:
zoom with new centring was not being called when zooming with the keyboard/menu button.

### DIFF
--- a/tilia/ui/timelines/collection/view.py
+++ b/tilia/ui/timelines/collection/view.py
@@ -2,13 +2,19 @@ from __future__ import annotations
 
 from PyQt6.QtCore import Qt
 from PyQt6.QtWidgets import QGraphicsView, QAbstractSlider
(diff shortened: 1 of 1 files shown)


FelipeDefensor (approved):
Ok. Let's make sure to add this in the next major version.

azfoo:
> Ok. Let's make sure to add this in the next major version.

Are you sure this doesn't go into the current version too? The non-modifier version is also affected. Zooming by the keyboard or the toolbar will cause centring to happen at some random point.

FelipeDefensor:
> > Ok. Let's make sure to add this in the next major version.
> 
> Are you sure this doesn't go into the current version too? The non-modifier version is also affected. Zooming by the keyboard or the toolbar will cause centring to happen at some random point.

Oh, yes, you're right. Glad you went ahead and merged it.

PR #193, opened 2024-09-20
Initial work on v0.5

FelipeDefensor:
Main changes:

- Greatly increases test performance by using a headless QT for testing;
- Fixes some CLI commands;
- Enable CLI tests again;
- Refactor media loading.

These changes are meant to be release in v0.5, and were waiting for the v0.4.3 release to be merged into the dev branch.


FelipeDefensor:
@azfoo, I know this is a lot, so don't worry about reviewing all of it. You can just glance at the commits to be aware of what changed and try to get any abvious mistakes.

FelipeDefensor:
You will notice some changes to the `script generate` command, as I ran into some conflicts when merging other changes to the `script` command. I chose to have a dedicated `script_generate` command instead of sharing the `script` parser with a `run` command. That makes for shorter commands: `script <path>` vs. `script run --path <path>` and `script_generate <path>` vs. `script generate --path <path>. We could absolutely have stayed with what we had, but I had trouble conciliating that with the other changes.

I removed the ScriptRunner class, and used the existing `tilia.ui.script.run` function instead, avoiding the extra code. I don't like the idea of having classes that hold no or very little state, as they can usually be simply replaced by functions, resulting in simpler code most of the time. I'm guilty of ignoring that with the `MediaLoader` class, but I also refactored it out of existence in this PR. Also, instantiating classes and having them (or any other object, for that matter) as global properties of a module can make testing harder. Again, I'm guilty of doing that in the `settings` module, so that's a refactor for later.

PR #196, opened 2024-10-10
Add Playback times

### LINKED ISSUE #178: Enable customization of effective media boundaries

azfoo:
#180, closes #178.
WIP:
~~fix CLI dialog~~


azfoo:
@FelipeDefensor, need a little help on this one: tests.ui.cli.test_load_media
in trying to avoid the dialog popping up on the CLI, I've called Get.WINDOW_STATE and checked for a NoReplyToRequest (because it is only served in QtUI and not the CLI), which works with the actual CLI, but fails on the test. Somewhere in the setup of the test, QtUI is initialised and is replying. Creating a new Get.UI_TYPE also fails because both UIs give conflicting responses. Any ideas on how to solve this?

FelipeDefensor:
> @FelipeDefensor, need a little help on this one: tests.ui.cli.test_load_media in trying to avoid the dialog popping up on the CLI, I've called Get.WINDOW_STATE and checked for a NoReplyToRequest (because it is only served in QtUI and not the CLI), which works with the actual CLI, but fails on the test. Somewhere in the setup of the test, QtUI is initialised and is replying. Creating a new Get.UI_TYPE also fails because both UIs give conflicting responses. Any ideas on how to solve this?

Will look into it.

FelipeDefensor:
I wasn't able to replicate the issue. `test.ui.cli.test_load_media` passes just fine here.

In general, to "patch" a dialog interaction, you should use the `tests.mock.Serve` context manager and pass it the desired response. Note that that will usually be a tuple with the "return value" and a boolean indicating wheter the dialog was succesful (that is, if it was not not closed or cancelled).

FelipeDefensor:
> I wasn't able to replicate the issue. `test.ui.cli.test_load_media` passes just fine here. Edit: I was running only `test.ui.cli.test_load_media.test_load_media`, now I am able to replicate.
> 
> In general, to "patch" a dialog interaction, you should use the `tests.mock.Serve` context manager and pass it the desired response. Note that that will usually be a tuple with the "return value" and a boolean indicating wheter the dialog was succesful (that is, if it was not not closed or cancelled).

FelipeDefensor:
Ok, so this is entirely on me. Contrary to what you would expect, the `marker_tl` fixture requests the `marker_tlui` fixture that request the `qtui` fixture. This is first-class spaghetti that I need to unravel. 

There were improvements to the tangled fixtures in v0.4.3, but apparently they were'nt enough. I will see if I can provide a permanent solution in the next few days. You can wait for that, or skip those tests with `@pytest.mark.skip` if you want to move on with this PR.

azfoo:
Marked skip for now. 
Do review if you have time - don't think there's much change in the code apart from the merge/rebase.

FelipeDefensor:
> There were improvements to the tangled fixtures in v0.4.3, but apparently they were'nt enough. I will see if I can provide a permanent solution in the next few days. You can wait for that, or skip those tests with @pytest.mark.skip if you want to move on with this PR. Let me know.

Easier than I thought, so I already pushed a commit to dev dealing with this.

Will review this PR later today.

FelipeDefensor:
Anne, sorry for the delay on the review. I was part of a congress today that took longer than I expected. Will review tomorrow, guaranteed.

tilia/media/player/base.py
```diff
@@ -176,8 +190,8 @@ def stop(self):
 
     def on_loop_changed(self, start_time, end_time):
         self.is_looping = start_time != end_time
-        if end_time == get(Get.MEDIA_DURATION):
-            if start_time != 0:
+        if end_time == self.playback_time.end:
+            if start_time != self.playback_time.start:
                 end_time -= self.E
```
FelipeDefensor, line 195: Did not get this. A comment clarifying would probably be nice.

tilia/media/player/base.py
```diff
@@ -352,3 +375,33 @@ def _engine_loop(self, is_looping: bool) -> None:
...
+
+    @end.setter
+    def end(self, value: float):
+        if value < self._start:
+            raise ValueError("End time must be greater than start time.")
+        self._end = value
+
+    @property
+    def duration(self):
+        return self._end - self._start
```
FelipeDefensor, line 407: Nice job with this one. If we find ourselves using this pattern often, we may want to use `pydantic`, which provides built-in validation for dataclasses.

tilia/timelines/base/timeline.py
```diff
@@ -193,11 +193,11 @@ def clear(self):
     def delete(self):
         self.component_manager.clear()
 
-    def scale(self, factor: float) -> None:
-        self.component_manager.scale(factor)
+    def scale(self, factor: float, offset_old: float, offset_new: float) -> None:
+        self.component_manager.scale(factor, offset_old, offset_new)
```
FelipeDefensor, line 197: I'm not sure I get whats going on here. Could you add a docstring explaining what the offsets do?

tilia/timelines/beat/timeline.py
```diff
@@ -380,11 +380,23 @@ def _validate_component_creation(
...
+        elif playback_time.end == 0 and time > media_duration:
+            return (
+                False,
+                f"Time '{time}' is greater than current media time '{media_duration}'",
+            )
+        elif time < playback_time.start:
+            return (
+                False,
+                f"Time '{time}' is less than current media start time '{playback_time.start}",
+            )
```
FelipeDefensor, line 399: We should abstract this into PointTimelineComponent when those commits meet.

tilia/timelines/hierarchy/timeline.py
```diff
@@ -115,16 +116,22 @@ def __init__(self):
...
         self, _, start: float, end: float, *args, **kwargs
     ):
+        playback_time = get(Get.MEDIA_TIMES_PLAYBACK)
         media_duration = get(Get.MEDIA_DURATION)
-        if start > media_duration:
+        if playback_time.end != 0 and (
+            end > playback_time.end
+            or start > playback_time.end
+            or end < playback_time.start
+            or start < playback_time.start
```
FelipeDefensor, line 125: The `end > start` is handled below, so we should only check if `start < playback_time.start or end > playback_time.end`.

tilia/timelines/hierarchy/timeline.py
```diff
@@ -115,16 +116,22 @@ def __init__(self):
...
+        ):
             return (
                 False,
-                f"Start time '{start}' is bigger than media time '{media_duration}'",
+                f"Time '{start} - {end}' is out of current media time bounds '{playback_time.start} - {playback_time.end}'",
             )
-        elif end > media_duration:
+        elif playback_time.end == 0 and (end > media_duration or start < 0):
```
FelipeDefensor, line 131: What is this checking for that is not handled by the previous cases? A comment would be nice.

tilia/timelines/hierarchy/timeline.py
```diff
@@ -601,20 +608,28 @@ def delete_component(self, component: Hierarchy, **kwargs) -> None:
...
-                    self.post_component_event(Post.HIERARCHY_POSITION_CHANGED, hrc.id)
+            if (hrc.end > end and hrc.start > end) or (
+                hrc.start < start and hrc.end < start
+            ):
+                self.delete_component(hrc)
+            elif hrc.end > end and hrc.start >= start:
+                hrc.end = end
+                if hrc.post_end > end:
+                    hrc.post_end = end
+                self.post_component_event(Post.HIERARCHY_POSITION_CHANGED, hrc.id)
```
FelipeDefensor, line 627: We should not have to manually notify about changes to a component's attributes. This was already wrong in my previous code. Actually, `Post.HIERARCHY_POSITION_CHANGED` does nothing, as no one is listening for it. That and dot-assingning component's attributes should raise eyebrows.

What we should do is use `self.set_component_data` which posts the correct events after doing the assignment.

tilia/ui/timelines/audiowave/timeline.py
```diff
@@ -64,8 +64,9 @@ def on_side_arrow_press(self, side: Side):
...
 
     def get_inspector_dict(self):
-        start_time = get(Get.MEDIA_DURATION)
-        end_time = 0
+        playback_times = get(Get.MEDIA_TIMES_PLAYBACK)
+        start_time = playback_times.end
+        end_time = playback_times.start
         for element in self.selected_elements:
             if element.get_data("start") < start_time:
                 start_time = element.get_data("start")
```
FelipeDefensor: I'm just realizing this is O(n) and given that we tipically have thousands of AmplitudeBars, this is pretty bad. No wonder there's lag when box selecting AmplitudeBars. Let's fix this in another commit ASAP.

tilia/ui/timelines/collection/collection.py
```diff
@@ -683,11 +683,11 @@ def on_loop_ignore_delete(self, tl_id: int, comp_id: int):
     def loop_cancel(self):
         self.loop_elements.clear()
         post(Post.PLAYER_UI_UPDATE, PlayerToolbarElement.TOGGLE_LOOP, False)
-        self.on_loop_change(0, 0)
+        self.on_loop_change(self.selected_time, self.selected_time)
```
FelipeDefensor, line 686: Why this?

tilia/ui/windows/metadata.py
```diff
@@ -109,11 +112,60 @@ def populate_with_read_only_fields(self):
...
             value_label = QLabel(value)
             value_label.setWordWrap(True)
             self.form_layout.addRow(QLabel(name.capitalize()), value_label)
 
+    def populate_with_media_times(self):
+        absolute_time = get(Get.MEDIA_TIMES_ABSOLUTE)
+        playback_time = get(Get.MEDIA_TIMES_PLAYBACK)
+
+        if absolute_time.start == 0.0 and absolute_time.end == 0.0:
+            self.has_times = False
```
FelipeDefensor, line 125: Make this a guard clause to avoid the following long indented block.

tilia/ui/windows/metadata.py
```diff
@@ -109,11 +112,60 @@ def populate_with_read_only_fields(self):
...
+            self.end_time.setSuffix(" s")
+            self.end_time.setKeyboardTracking(False)
+            self.end_time.setSingleStep(1.0)
+            self.end_time.valueChanged.connect(lambda value: on_end_time_changed(value))
+
+            self.actual_duration = QLabel(
+                format_media_time(playback_time.duration, False)
+            )
+
+            self.has_times = True
```
FelipeDefensor, line 151: Move this to the top to be closer to the contasting statement on the guard clause case.

tilia/ui/windows/metadata.py
```diff
@@ -109,11 +112,60 @@ def populate_with_read_only_fields(self):
...
+            )
+
+            self.end_time = QDoubleSpinBox()
+            self.end_time.setMaximum(absolute_time.end)
+            self.end_time.setMinimum(playback_time.start)
+            self.end_time.setValue(playback_time.end)
+            self.end_time.setSuffix(" s")
+            self.end_time.setKeyboardTracking(False)
+            self.end_time.setSingleStep(1.0)
+            self.end_time.valueChanged.connect(lambda value: on_end_time_changed(value))
```
FelipeDefensor, line 145: Do you think abstracting the duplicated code into a class is worth it in this case?

tilia/ui/dialogs/scale_or_crop.py
```diff
@@ -0,0 +1,95 @@
...
+from enum import Enum
+
+from tilia.ui.format import format_media_time
+
+
+class ScaleOrCrop(QDialog):
+    class ActionToTake(Enum):
+        SCALE = 0
+        CROP = 1
+        PROMPT = 2
```
FelipeDefensor, line 21: I'm not opposed to using an enum instead of strings, but:
- This should be a `StrEnum` to increase readability (use `enum.auto()` for the values instead of hard-coding`
- We should use the `enum` everywhere. There are still places where Literal['yes', 'no', 'propmpt'] is being used.
- This should be in a "CLI-friendly" place. There should be no reason for the CLI to use code in the `ui.dialogs` dir.

tilia/ui/timelines/audiowave/timeline.py
azfoo: Quick question: is the list of selected elements always sorted?

tilia/app.py
```diff
@@ -124,29 +144,46 @@ def get_id(self) -> int:
...
+                self.cached_media_end,
+                playback_time.start,
+                playback_time.end,
+            )
+            if not success:
+                return
+        else:
+            scale_or_crop = self.should_scale_timelines
+
+        match scale_or_crop:
```
FelipeDefensor, line 175: Can't we just `match self.should_scale_timelines`?

tilia/app.py
```diff
@@ -164,9 +201,11 @@ def prompt_crop_timelines(self):
     def _check_if_media_exists(path: str) -> bool:
         return re.match(tilia.constants.YOUTUBE_URL_REGEX, path) or Path(path).exists()
 
-    def _setup_file_media(self, path: str, duration: float | None):
+    def _setup_file_media(
+        self, path: str, duration: float | None, start: float, end: float
+    ):
         if duration:
-            self.set_file_media_duration(duration)
+            post(Post.FILE_MEDIA_DURATION_CHANGED, duration, start, end)
```
FelipeDefensor, line 208: Those seem to do different things. Was there anything wrong before?

FelipeDefensor (changes requested):
The code is pretty clean and everything seems to be working. I tried to be more thorough than normal in the review as experience (or trauma) tells me that media loading bugs tend to be hard to spot.

I will try to write a billion tests for this. In the meantime, there are some minor issues to be adressed.

Good job!

tilia/ui/timelines/audiowave/timeline.py
FelipeDefensor: Yes, always sorted, despite the name `ElementManager._add_to_selected_elements_set`.

tilia/ui/dialogs/scale_or_crop.py
FelipeDefensor: I had not realized how you were handling the CLI case in `ScaleOrCrop.select`, sorry for that. Some issues with that:

- I feel like we should avoid cathing `NoReplytoRequest` to branch the code. I always imagined that exception as just a wrapper around `ValuError` that made things more understandable.
- That makes the CLI dependent on `QDialog`, which increases coupling between the CLI and `PyQt`.
- It strikes me as an instance of too many responsabilites for a function, that is, handling user interaction in two types of user interfaces.

Please tell me if you feel differently on those issues.

If you don't mind, I would like to commit a refactor that separates the GUI and CLI dialog.

tilia/app.py
azfoo: I've chosen not to set `self.should_scale_timelines` so that it doesn't need to be unset after this operation is done.

tilia/app.py
azfoo: Not calling the post means the player doesn't get updated. Currently, any changes made to the media duration calls `Post.FILE_MEDIA_DURATION_CHANGED` which gets the player to set up the necessary playback times. Player then calls `Post.PLAYER_DURATION_AVAILABLE` to update the rest of the app.

tilia/timelines/hierarchy/timeline.py
azfoo: Fair point. I think I left this over from before I made changes to how files were loaded.  It should be fine without that clause.

tilia/ui/timelines/collection/collection.py
azfoo: To avoid drawing a (empty) loop box beyond the playback time range, since playback_start could be more than 0.

tilia/ui/dialogs/scale_or_crop.py
azfoo: The values are hard coded so I can check which of the radio buttons in the dialog have been selected. 
The Literals in the App class are all from the CLI that are also converted into Enums - will move the logic into the CLI. 
Will also separate the logic between the GUI and CLI.

tilia/ui/windows/metadata.py
azfoo: think its fine for now. am thinking of making a range slider  to replace this at some point (as mentioned previously)

tilia/app.py
FelipeDefensor: Ok, got it.

tilia/ui/windows/metadata.py
FelipeDefensor: Oh, yes, forgot about that. All good, then.

FelipeDefensor (changes requested):
The changes are perfect :).

Few new bugs I found:
- Pre-start and post-end were not being scaled correctly. I fixed that with a commit.
- Something goes wrong shen scaling multiple segmentlikes. I added a test for that behavior.
- Undoing a media load with scale or crop triggers the dialog again. We could either reset the undo history when loading a new media or somehow store whether the user chose the scale or crop options.

We are almost there!

PR #207, opened 2024-10-18
AudioWave inspector remove search

azfoo:
fix: remove redundant search
selected_elements is already sorted


FelipeDefensor (approved):
All good.

PR #208, opened 2024-10-20
Abstract component validation

FelipeDefensor:
Now that we have the `PointLikeTimelineComponent` and `SegmentLikeTimelineComponent`, this abstraction could be more easily located.

I also added some more validation to hierarchies. That should help spotting bugs earlier.


azfoo (changes requested):
This could be further abstracted. Since each `_validate_component_creation` is composed of a `validate_<something>like_component_creation` and a `validate_unique_position_<something>like_component`, `_validate_component_creation` could be put into the parent class with each of the two child validates marked not initialised (to initialise in the child class).

i.e.`_validate_component_creation` goes into `TimelineComponentManager`, calls `component_kinds.validate_component_time` and `component_kinds.validate_unique_position`, so the two validators can be in the respective `TimelineComponent` classes (and probably renamed to `validate_component_time` and `validate_unique_position` or something similar)

tilia/timelines/base/component/base.py
```diff
@@ -0,0 +1,70 @@
...
+        self.id = id
+
+    def __str__(self):
+        return get_tilia_class_string(self)
+
+    def __lt__(self, other):
+        return self.ordinal < other.ordinal
+
+    @property
+    def user_friendly_name(self):
```
azfoo, line 34: could do with a better name - this sounds like the original name is not readable. Maybe frontend_name or name_alt?

FelipeDefensor: Felt the same. I like frontend_name.

FelipeDefensor:
> This could be further abstracted. Since each `_validate_component_creation` is composed of a `validate_<something>like_component_creation` and a `validate_unique_position_<something>like_component`, `_validate_component_creation` could be put into the parent class with each of the two child validates marked not initialised (to initialise in the child class).
> 
> i.e.`_validate_component_creation` is moved into `TimelineComponentManager`, calls `self.component_kinds.validate_component_time` and `component_kinds.validate_unique_position`, so the two validators can be in the respective `TimelineComponent` classes (and probably renamed to `validate_component_time` and `validate_unique_position` or something similar)

Good point. Will do that.

azfoo (approved):

PR #211, opened 2024-10-21
fix: args typo

azfoo:
was causing pdf & beat to not load

### DIFF
--- a/tilia/ui/timelines/beat/timeline.py
+++ b/tilia/ui/timelines/beat/timeline.py
@@ -27,19 +27,23 @@ class BeatTimelineUI(TimelineUI):
         "measure_numbers",
         "beats_that_start_measures",
         "measures_to_force_display",
-        "beats_in_measure"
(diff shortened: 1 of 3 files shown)


FelipeDefensor (approved):
Glad you caught this right away. Merge this ASAP.

FelipeDefensor:
Important lessons to learn here:
1. Using `Serve` to mock user input may conceal bugs. We should find a way to simulate user interactions more faithfully, in those cases. Using `QDialog.show` instead of `QDialog.exec` with `qtbot` might be the way, as we have previsously discussed.
2. While we don't have that in place, we should always test things manually at least once before pushing changes.

azfoo:
That was a big oops. Should have caught this when I tested.

FelipeDefensor:
> That was a big oops. Should have caught this when I tested.

Big oops for both of us, I should never had pushed it without testing it myself.

PR #214, opened 2024-10-24
Use <enter> to toggle Inspector

### LINKED ISSUE #203: Convenient way needed to open and close "Inspect" menu (or rather, to label a component)
**Desired behaviour**

Double-click on the component of a hierarchy or use a hotkey to open the "Inspect" mask in order to label it and/or assign a function or type. Press `Enter` or `Esc` to close it.

**Current behaviour**

No hotkeys are assigned. One needs to right-click and select "Inspect". The only way to close the mask is using the mouse (or the OS's hotkey to close the window which is not suited for a "confirm" action).

**Discussion**

Alternatively, and maybe even more desirable, the user could simply start typing to label a component (which would require adding `Ctrl/Cmd` to the hotkeys which are currently single-key). A good example for this straightforward way of labelling things is https://draw.io Considering the importance of labels in an annotation app, I reckon it would be worth the change. A streamlined way to annotate with keyboard only is would be a killer feature IMHO.

azfoo:
closes #203

### DIFF
--- a/tilia/ui/qtui.py
+++ b/tilia/ui/qtui.py
@@ -30,7 +30,14 @@ from .dialogs.by_time_or_by_measure import ByTimeOrByMeasure
 from .dialogs.crash import CrashDialog
 from .menubar import TiliaMenuBar
 from tilia.ui.timelines.collection.collection import TimelineUIs
-from .menus import TimelinesMenu, HierarchyMenu, MarkerMenu, BeatMenu, HarmonyMenu, PdfMenu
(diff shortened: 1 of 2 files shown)


FelipeDefensor (approved):
Looks good to me. I like the idea of using <enter> *both* to open and close. Additionally, I think we should still have <Esc> as an option to close.

azfoo:
docs seem to say that `esc` closes, but that doesn't happen. will reimplement then merge.

PR #216, opened 2024-10-24
Make zooming a lot faster

FelipeDefensor:
This implements a lot of small optimizations that add up to a ~20x speed increase in zooming. This makes zooming on large files, previously prohibitively slow, smooth enough to not be bothersome.

The main optimizations were:
- Passing `get_data` and `set_data` functions to `TimelinUIElement` instances so they have faster access to component properties
- Creating a `TimeXConverter` class to cache values (such as media duration) used in almost every `TimelineUIElement` position caculation
- Remove duplicate calculation of positions values on specific `TimelineUIElement` subclasses (e.g. `HierarchyUI`
- Optmizing `HierarchyUI.update_lavel`
- Cache settings, as some are acessed every time a component needs to be repositioned

### DIFF
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -8,7 +8,7 @@ import pytest
 import tests.utils
 from tests.mock import Serve, PatchPost
 from tilia.media.player import YouTubePlayer, QtAudioPlayer
-from tilia.settings import settings_manager
+from tilia.settings import settings
 
 from tilia.requests import Get, Post, post
 from tilia.ui.actions import TiliaAction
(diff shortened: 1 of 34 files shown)


FelipeDefensor:
As far as I can remember, this had a bug where the components could become misaligned after zooming. Rebasing the most recent commits from `dev` seems to have fixed it, I was not able to reproduce the bug after that. I would be extra cautious when testing this one, tough.

azfoo (approved):
I think that fixes the last of the issues we had before the rebase.

FelipeDefensor:
Indeed, it's better to keep the in the same class as the manager. I will write some tests for opening and closing the windows, that should be something we check with every commit.

Issue #205, opened 2024-10-15
Bug: Beat Layer no longer displayed + unable to save and close window after opening file


m-k94:
I was working on a file with different kinds of annotations and a beat/measure layer for a few days and everything was normal. But now when I open it, I just see the harmony and form annotations, but not the beat and marker layers that I had before. When I open the TLA file in an editor, all the information appears to still be there. And it even does get properly displayed in the Web-App: https://tilia-ad98d.web.app/viewer/66 

It also has some other strange side effects on the desktop version: Once the file is opened, I can interact with it, but I am unable to save it, open a different file or close the window normally. I have to do use task manager. Other files I created - so far - do not cause such issues.

I _think_ the bug might have started when I pressed Ctrl+z to undo an action. That unexpectedly caused a lot of stuff to disappear all at once, and when I reopened everything to start over from the last save, it started behaving as described above.

FelipeDefensor:
Could you send me the .tla?

I'm guessing that something went wrong when undoing leading to an invalid timeline state. The overhaul that Loading a file in the web app happens slightly differently, so the faulty state may not have consequences there.

I would imagine that downloading the file you uploaded into the web app should give you a healthy file.

FelipeDefensor:
The undoing/redoing process already went through an overhaul in the `optimize-restore-state` branch, so the changes there might have prevented whatever lead to the faulty state in the first place.

In any case, the app should be resilient to things going wrong when undoing/redoing. Perhaps it could check if the timeline state after the restore is equal to the desired state (the previous one, when undoing), and, if not, revert to the last known healthy state. Will look into that for v0.5.

m-k94:
Thanks for the quick response! I tried downloading the file, but it causes the app to fully crash:

![image](https://github.com/user-attachments/assets/653b0c08-3730-4793-af5b-3f74f402a6ea)

Here is my original file: https://drive.google.com/file/d/1FswDATSIgQifnoTLTHWJT6FZfSptczV4/view?usp=sharing

FelipeDefensor:
I managed to restore the file to valid state. Please check if you can open this: [Bruch Skizze_recovered.zip](https://github.com/user-attachments/files/17384206/Bruch.Skizze_recovered.zip)

m-k94:
That file works fine, thanks!

johentsch:
Ran into the same problem with the same piece. Cannot recall the exact sequence of events but I did store these two error messages:

```python-traceback
[...]

PR #215, opened 2024-10-24
Make restoring state a lot faster

FelipeDefensor:

PR #223, opened 2024-11-04
Recover to previous state when user actions or undoing/redoing raises an exception

FelipeDefensor:
This should prevent a lot of crashes, while still outputting a useful error message. Additionally, this makes debugging easier as the previous (healthy) state can be saved and debugged later.

Although we have not found the exact bug that lead to invalid state at #205 being created, this would have prevented the file from going to an invalid state with an undo.


azfoo:
will review in a little bit. 
One thing that I've noticed is that sometimes data validation causes elements to not load correctly.

Eg. in 37ea1f0c3ed2dd69834c89bc9e847769e90f07be, the reason for failure is that the previous elements that have not yet been updated don't allow the current edit to take place. 

I think the validation process might need to be updated at some point, where validation of individual components doesn't take place until the entire edit process has been completed. That might solve bugs that involve editing multiple components at once (like undo/redo and crop/scale)

FelipeDefensor:
> I think the validation process might need to be updated at some point, where validation of individual components doesn't take place until the entire edit process has been completed. That might solve bugs that involve editing multiple components at once (like undo/redo and crop/scale)

Fair. Maybe the validation is checking for things that would be better handled by tests. Perhaps we should only validate user input (e.g. attributes from imported CSVs) to have more freedom when creating\editing things ourselves.

azfoo:
> > I think the validation process might need to be updated at some point, where validation of individual components doesn't take place until the entire edit process has been completed. That might solve bugs that involve editing multiple components at once (like undo/redo and crop/scale)
> 
> Fair. Maybe the validation is checking for things that would be better handled by tests. Perhaps we should only validate user input (e.g. attributes from imported CSVs) to have more freedom when creating\editing things ourselves.

Validation is important and shouldn't be skipped. 
Validating at the end of a process, instead of at every step, would also reduce computation time.

FelipeDefensor:
> Validation is important and shouldn't be skipped. Validating at the end of a process, instead of at every step, would also reduce computation time.

If we can take the performance hit, as has been the case, I'm not necessarily opposed to keep validating some of the internal inputs. In any case, I agree that we should rethinkg it to avoid the problems you pointed out.

FelipeDefensor:
@azfoo why are we closing this?

azfoo:
my bad. that was an accident and I couldn't find it to restore.

tilia/ui/timelines/collection/request_handler.py
```diff
@@ -50,8 +48,14 @@ def on_timeline_add(self, kind: TimelineKind):
                 return
             kwargs |= additional_args
 
-        self.timelines.create_timeline(kind=kind, components=None, name=name, **kwargs)
+        timeline = self.timelines.create_timeline(
+            kind=kind, components=None, name=name, **kwargs
+        )
+
+        return bool(timeline)
```
azfoo, line 55: is there anything that checks this response?

tilia/app.py
```diff
@@ -160,12 +173,42 @@ def load_media(self, path: str, record: bool = True, scale_timelines: Literal['y
...
+        """
+        Clears the app and attemps to restore the given state.
+        Unlike `on_restore_state` this will crash if an error occurs during
+        the restoration.
+        This is meant to be used after an exception occurred, so if the
+        restoration fails, we are likely in an invalid state and
+        therefore crashing is the best option.
+        """
+        self.on_clear()
+        self._restore_app_state(state)
```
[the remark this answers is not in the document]
FelipeDefensor: To prevent a second attempt we simply run `self._restore_app_state` without catching exceptions. If any are raised, then the program crashes.

There's no hash of the entire state, only of its timelines, and that is not used for this particular purpose.

tilia/ui/timelines/collection/request_handler.py
FelipeDefensor: `TimelineUIs.on_timeline_uis_request` uses this value to determine if it should record the state in the UndoManager (in case of success) or to display an error message (in case of failure).

`bool(timeline)` makes no sense here, though. It should simply return `True` if it reaches that point or `False` in the previous early returns.

azfoo (approved):

FelipeDefensor:
Waiting for approval on #215 to merge this.

FelipeDefensor:
I made some mistakes when merging this to the target branch which made the rebase very difficult. I cherry-picked the commits from this branch into dev to get around that, so this won't show as merged. I tried doing this https://stackoverflow.com/a/43509991/15862653, but it also didnt work.

PR #227, opened 2024-11-05
Make it possible to create timeline with no media loaded

### LINKED ISSUE #218: Allow for annotating beat grid without media
**Desired behaviour**

Open TiLiA, create beat grid, start annotating.

**Current behaviour**

Adding beat grid fails with `Cannot create timeline with no media loaded`.

**Discussion**

This is probably part of a bigger issue which is lacking convenience in creating and editing beat grids. Like, when someone enters a grid as numbers, it could be generated right away by distributing beats evenly (as a reasonable default). In my view, the whole functionality of aligning audio with symbolic timelines is huge and involves quite a few slippery slopes, but on the other hand, good solutions already exist for many of them (e.g. in the context of the [match format]([https://doi.org/10.17613/37mm-pt05](https://hcommons.org/deposits/item/hc:59273))). I saw that you can import the beatgrid from CSV so probably you have already thought about creating compatibility with the Sonic Visualizer?

I think this topic warrants important conceptual work for the planned inclusion of scores.

FelipeDefensor:
The user will now be asked whether to load a media file or to set the media duration manually when trying to create a timeline with no media loaded. This saves a few clicks in the former but, more importantly, it gives the option of annotating with no media loaded. This will be crucial if the user wants to work directly with a score in a score timeline.

I also refactored the `Get.FROM_USER_<SOMETHING>` responses to more consistenly return something of the form (success: bool, value: Any).

Closes #218.


azfoo (changes requested):
hierarchy handles causes crash when dragged.

azfoo:
> hierarchy handles causes crash when dragged.

Actually, no. This seems to be a bug introduced before this.

FelipeDefensor:
I fixed the dragging in the dev branch. I merged it (instead of rebasing) to avoid having to force push.

azfoo (approved):

PR #228, opened 2024-11-12
Add action to fill beat timeline with evenly-spaced beats

FelipeDefensor:


FelipeDefensor:
Started working from wrong branch. Will redo.

PR #229, opened 2024-11-13
Populate beat timeline with evenly-spaced beats

FelipeDefensor:
Further addresses #218.


azfoo (commented):
Should there be a warning about deleting existing beats?

FelipeDefensor:
> Should there be a warning about deleting existing beats?

Yes, most likely. Will add that.

FelipeDefensor:
> Should there be a warning about deleting existing beats?

Glad you found this, because I also forgot to actually delete the existing beats.

PR #234, opened 2024-11-21
More precise BeatTimeline.get_times_by_measure

FelipeDefensor:
Take beat times into consideration in BeatTimeline.get_times_by_measure.

### DIFF
--- a/tests/timelines/beat/test_beat_timeline.py
+++ b/tests/timelines/beat/test_beat_timeline.py
@@ -379,6 +379,37 @@ class TestBeatTimeline:
 
         assert beat_tl.get_time_by_measure(-1) == []
 
+    def test_get_times_by_measure_measure_zero(self, beat_tl):
+        beat_tl.set_data("beat_pattern", [2])
+        beat_tl.create_beat(time=5)
+        beat_tl.create_beat(time=6)
+        beat_tl.create_beat(time=7)
+
+        beat_tl.recalculate_measures()
+
+        assert beat_tl.get_time_by_measure(0) == [3]
+        assert beat_tl.get_time_by_measure(0, 0.5) == [4]
+        assert beat_tl.get_time_by_measure(0, 0.75) == [4.5]
+
+    def test_get_times_by_measure_measure_zero_has_zero_as_minimum(self, beat_tl):
+        beat_tl.set_data("beat_pattern", [2])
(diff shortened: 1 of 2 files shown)


FelipeDefensor:
@azfoo hold up a bit on this review, I just remembered something.

FelipeDefensor:
Done, you can go ahead.

azfoo:
Sorry I forgot about the existence of this and accidentally did work for this in the mxl branch (which I separated out to 
[this branch](https://github.com/azfoo/TiLiA/tree/better-beat). Haven't written tests for it, but this creates/uses a lookup table which would be useful for quick conversion to score positions in the mxl-parser. (There's definitely room to make the lookup table smaller if necessary.) There shouldn't be an issue with measure zero and the last measure in this one, except you'll need to put in an extra beat at the end of the last measure (belonging to last + 1 measure) to end things off properly.

FelipeDefensor:
> Sorry I forgot about the existence of this and accidentally did work for this in the mxl branch (which I separated out to [this branch](https://github.com/azfoo/TiLiA/tree/better-beat). Haven't written tests for it, but this creates/uses a lookup table which would be useful for quick conversion to score positions in the mxl-parser. (There's definitely room to make the lookup table smaller if necessary.) There shouldn't be an issue with measure zero and the last measure in this one, except you'll need to put in an extra beat at the end of the last measure (belonging to last + 1 measure) to end things off properly.

No problem. So I imagine you version is more performant. Can we use the tests I wrote also in that branch?

azfoo:
> No problem. So I imagine you version is more performant. Can we use the tests I wrote also in that branch?
Tests added and passed. 
I removed measure zero because it shouldn't be an issue if the user manually adds in a measure 0. Having measure zero without an actual user-defined start point would be too arbitrary.

FelipeDefensor:
> > No problem. So I imagine you version is more performant. Can we use the tests I wrote also in that branch?
> > Tests added and passed.
> 
> I removed measure zero because it shouldn't be an issue if the user manually adds in a measure 0. Having measure zero without an actual user-defined start point would be too arbitrary.

Not sure I agree. I think most users will be surprised that anacrusis don't get magically imported, even with no explicit setup on their part, but I'm fine with not doing that for now and waiting for user feedback.

FelipeDefensor:
Closing this in favor of the more performant version mentioned above.

PR #239, opened 2024-12-16
Some fixes and improvements to score timeline

FelipeDefensor:
The three main things:
- Moving SVGViewer to frontend
- Prompting for the creation of measure zero when importing
- Fix wrong barline position after importing

The specific commit messages go into more detail. This also has some other commits for other smaller useful things.


azfoo (commented):
The reason for leaving it in the back end is CLI usability for saving SVG data. mxl data is converted to SVG data through an invisible web channel and saved from there. What could happen instead is separating the mxl-svg conversion from the display then separating front and backend from there. Will work on that.

FelipeDefensor:
> What could happen instead is separating the mxl-svg conversion from the display then separating front and backend from there. Will work on that.

If you did get a chance to review rest of this PR, we could merge this commit and make a new PR for those changes.

tilia/parsers/score/musicxml.py
```diff
@@ -315,7 +315,15 @@ def _parse_staves(tree: etree.ElementTree):
     if tree.tag == "score-timewise":
         tree = _convert_to_partwise(tree)
     elif tree.tag != "score-partwise":
-        raise ValueError(f"File `{path}` is not valid musicxml.")
+        return False, [f"File `{path}` is not valid musicxml."]
+
+    if tree.find('.//measure[@number="0"]') is not None and (0 not in beat_tl.measure_numbers):
```
azfoo, line 320: this is too generalised a solution - if the user does not actually want to import the entire score, not having a measure 0 would prevent the entire process from happening.

perhaps a dialog that shows which beat_tl measures are available and the measure number range found in the score instead so the user can confirm which measures are needed?

I'm still not terribly keen on creating measure 0 for the user - part of having user input timestamps is that the beats of music are not necessarily even and predictable. Resetting measure numbers or actually inputting a measure 0 is not too many clicks away.

tilia/ui/timelines/score/timeline.py
```diff
@@ -60,6 +61,9 @@ class ScoreTimelineUI(TimelineUI):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+
+        self.__svg_view = None
```
azfoo, line 65: doesn't work - will cause view to disappear whenever any kind of action is done on it. but will fix after this PR.

azfoo (changes requested):

tilia/parsers/score/musicxml.py
FelipeDefensor: > this is too generalised a solution - if the user does not actually want to import the entire score, not having a measure 0 would prevent the entire process from happening.

That is a fair point, although I imagine importing the entire score will be by far the most common use case. Stopping the import if no measure 0 is found is to prevent the crash that would follow due to the lack of a clef in measure 1 for positioning the beats. What we could do instead is to "push" the clefs to measure 1 when there is no measure 0. What do you think? Even in that case, I think we should prompt for adding measure 0, as I think most users will expect the anacrusis to be magically imported.

> perhaps a dialog that shows which beat_tl measures are available and the measure number range found in the score instead so the user can confirm which measures are needed?

This would be an extra step for all users to cover only a niche use case, namely, the user has beats for the whole media but only wants to import part of the score. If the user only wants the score for a portion of the media, then I would expect that they only create beats for that portion. In that case, "pushing" the clefs to the first imported beat would prevent crashes.

> I'm still not terribly keen on creating measure 0 for the user - part of having user input timestamps is that the beats of music are not necessarily even and predictable. Resetting measure numbers or actually inputting a measure 0 is not too many clicks away.

I think most users won't realize the anacrusis are missing beacuse there is no measure 0, they will simply think something is wrong with the import. I, for one, never realized anacrusis measure had a number in score editors until recently. I though of them as a special unumbered case. If the user wants to have precise timestamps for measure 0, then they can create it manually, that is also only a few clicks away, but I rather we add that extra step for what I think will be the minority of users. Maybe we disagree with what will be the standard use case and to which extent it will be standard.

tilia/ui/timelines/score/timeline.py
FelipeDefensor: Maybe that explains #246?

azfoo: yes. I've been working on a fix for that but will push after this one has been merged

tilia/parsers/score/musicxml.py
azfoo: "Pushing" clefs would be a necessary thing whenever an incomplete score is imported. Will add that on to the score-parser branch.

Still, implementing a forced measure 0 this way would seem to be a weird bug if the user does not import any measure close by. Expanding this solution for an anacrusis that happens in the middle of the score (which would not be numbered 0) would be a logical next step, but detecting the existence of anacruses might fall outside the scope of this project.

FelipeDefensor: > "Pushing" clefs would be a necessary thing whenever an incomplete score is imported. Will add that on to the score-parser branch.

What about we merge this and you make a new PR? That way avoid a potentailly large rebase of another branch.

> Still, implementing a forced measure 0 this way would seem to be a weird bug if the user does not import any measure close by.

Nice catch. To get arount that we could prompt for measure 0 only if measure 1 is being imported.

> Expanding this solution for an anacrusis that happens in the middle of the score (which would not be numbered 0) would be a logical next step, but detecting the existence of anacruses might fall outside the scope of this project.

Agreed. There's no easy way to detect an anacrusis in the middle of the score.

azfoo: right. could you implement the check for measure 0/1, then we merge and I'll start work.

FelipeDefensor: Done. Will merge.

PR #245, opened 2024-12-17
Prevent setting non-positive media duration

FelipeDefensor:
Min is necessary as non-positive values may lead to crashes and default is nice when the user (or the developer) just wants to create a timeline and does not care about the specific media duration.

I also generalized some ask_for_ functions to make this change easier.

### DIFF
--- a/tilia/ui/dialogs/basic.py
+++ b/tilia/ui/dialogs/basic.py
@@ -13,17 +13,17 @@ def ask_for_color(
 
 def ask_for_int(title: str, prompt: str, **kwargs) -> tuple[bool, int]:
     number, accepted = QInputDialog().getInt(None, title, prompt, **kwargs)
-    return bool(accepted) & True, number
+    return accepted, number
 
 
 def ask_for_string(title: str, prompt: str, **kwargs) -> tuple[bool, str]:
     string, accepted = QInputDialog().getText(None, title, prompt, **kwargs)
-    return bool(accepted) & True, string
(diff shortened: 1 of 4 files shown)


azfoo (approved):

tilia/ui/dialogs/basic.py
```diff
@@ -13,17 +13,17 @@ def ask_for_color(
 
 def ask_for_int(title: str, prompt: str, **kwargs) -> tuple[bool, int]:
     number, accepted = QInputDialog().getInt(None, title, prompt, **kwargs)
-    return accepted, number
+    return bool(accepted) & True, number
```
FelipeDefensor, line 16: Sorry, I didn't get this one. Why do we need this?

FelipeDefensor: @azfoo don't forget about this.

azfoo: oops. QInputDialog returns `bool | None`. I suppose it isn't entirely necessary but the operation ensures that a bool is actually returned. I'm happy for that commit to be removed.

FelipeDefensor: Oh, that would sense, although we would't need the `& True`, but `QInputDialog.getInt` does return an int. From https://doc.qt.io/qt-6/qinputdialog.html#getInt: "On success, this function returns the integer which has been entered by the user; on failure, it returns the initial value."

PR #249, opened 2024-12-18
Fix frontend names

FelipeDefensor:


azfoo (approved):
Not sure if this is an issue to be solved here, but the duplication error message doesn't come up for beat and marker components.

FelipeDefensor:
> Not sure if this is an issue to be solved here, but the duplication error message doesn't come up for beat and marker components.

Thanks for testing that. That's partly by design, as I thought the user will be able to figure out why the maker/beat is not being inserted in those cases. By that logic, PDF marker should do the same. Harmony/mode are a different case, as the add button triggers a dialog, and not the actual creation. I say we fix that later, as we already have our hands full.

PR #260, opened 2025-01-08
Fix varying inspector width

### LINKED ISSUE #253: Deselecting element changes inspector width and timeline view horizontal position
This makes the UI harder to read and may lead to unwanted behavior in some cases, as when dragging to create a selection box:

https://github.com/user-attachments/assets/8b5fff5d-2048-4650-bc62-e1c1aecea8da

FelipeDefensor:
I dindn't find a way to fix #253 without setting a fixed width to the inspector. Feel free to try it yourself, if you think it is worth it.


FelipeDefensor:
Closes #253

azfoo:
hold on with this one. going to try and see if there is a better solution.

FelipeDefensor:
> hold on with this one. going to try and see if there is a better solution.

Please do.

Issue #264, opened 2025-01-22
Display "Power Chords" as X5 instead of Xpower


m-k94:
When you select "Power" as a chord type, it gets displayed like this:

![Image](https://github.com/user-attachments/assets/87826605-d606-4929-9eff-38b32b3ccb2a)

Power chords are conventionally written as "A5", see for example [here](https://tabs.ultimate-guitar.com/tab/black-sabbath/iron-man-chords-3261149): 
I recommend adopting this way, so as to make the label shorter and easier to understand!

"Tristan" and "Pedal" have similar issues, but are much more rare. Maybe display these in brackets, so they don't get merged with the Note Letter, i.e. `A (Tristan)` instead of `Atristan`.

FelipeDefensor:
+1 for displaying power chords as X5.

"Pedal" is displayed as a subscript, so it is clearly distinct from the root letter:

![Image](https://github.com/user-attachments/assets/8e13c524-66c2-4a67-897c-acd7eb84bdf4)

I will shorten "pedal" to "ped." to make it more concise (the complete name will still be displayed in the inspector.

Tristan chords are currently displayed simply as "Tristan", without the root letter:

![Image](https://github.com/user-attachments/assets/dc485eba-1abb-4f25-924e-bce328515976)

I will add the root letter and make "tristan" a subscript.

m-k94:
Thanks!

In the current desktop version, pedals and tristans look like this for me:

![Image](https://github.com/user-attachments/assets/544acd52-4ef8-492e-aab3-e19d9c86535e)

but with subscript and abreviations it should be fine.

FelipeDefensor:
Oh, so we had already improved them a bit since the last release. Working on the fix right now, should make the PR later today.

FelipeDefensor:
If we are trying keep the logic of roman numeral notation, wouldn't a superscript 5 be the way to go? It is the only interval from the bass, after all, and we use superscripts when indicating a single interval, e.g. I^6.

PR #265, opened 2025-01-22
Improve readability of tristan, pedal and power harmonies symbols

FelipeDefensor:
Made "tristan" and "pedal" subscripts for chord symbols and abbbreviated them in chord symbols and roman numerals. Also used "X5" instead of "Xpower" for chord symbols of power chords. That is more conventional and more concise as pointed in #264. I opted to use a "pow" superscript as power chord suffix for roman numerals, as the "5" looks like an inversion suffix. 

Before:
image

After:
image


azfoo:
regarding power chords, classically understood, a chord needs at least 3 different notes so an open 5th power chord should not be considered a chord. X_(subscript)5 might be the better way to represent it if we fall back on to figured notation.

PR #282, opened 2025-02-04
Fix smooth scroll performance

### LINKED ISSUE #267: Slider getting out of synch with video in v0.5
One thing that has unfortunately gotten less reliable for me in the new version is the behaviour of the slider: I find myself having to restart the app a lot because after some time (about 10 minutes or so), it tends to be less and less synchonized with the actual time in the loaded video, until it is pretty much unusable. Sometimes the slider keeps moving after pausing the media, either for a few seconds or indefinitely. Or when I click on a segment, it jumps to a different place in the timeline or just lags. Since operations like splitting are done at the location of the slider, this can lead to some confusing annotation mistakes, that become apparent only after reopening the file.

I am not yet sure what causes this exactly, but at least for me it happens pretty reliably after some time of doing normal annotation work. Might be something to keep an eye on, see if this also happens for you when annotating a video for some time. On local files it looks good so far, though.

azfoo:
potentially closes #267.

These performance issues are likely related to the user's computer running out of memory. Setting this to default to off and at a more moderate value when it is on will likely reduce the number of such incidents.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -191,7 +191,7 @@ def use_test_settings(qapplication):
         constants_module.APP_NAME, f"DesktopTests"
     )
     settings_module.settings._check_all_default_settings_present()
-    settings_module.settings.set("general", "smooth-scroll", False)
+    settings_module.settings.set("general", "prioritise_performance", True)
     yield
 
 
--- a/tilia/settings.py
+++ b/tilia/settings.py
@@ -88,9 +88,7 @@ class SettingsManager(QObject):
     }
 
     def __init__(self):
-        self._settings = QSettings(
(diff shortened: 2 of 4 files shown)


FelipeDefensor:
Oh, any small changes you are working on or should we release v0.5.4? You can trigger the build yourself by tagging a commit with v0.5.4 (and then pushing the tag). I can fill you in the rest of the release proccess later.

FelipeDefensor:
I'm merging this to get it into v0.5.4, which I'm releasing now with the important #284. @azfoo let me know what you think about this:

> Approved, but I'm not sure we should rename the setting to prioritise_performance. That makes it sound like there are two correctly working behaviors that differ in performance, which doesn't seem to me to be the case, as there appears to be a bug in the smooth-scrolling code, maybe leading to a memory leak.

PR #283, opened 2025-02-05
fix: non-changeable inspector width

azfoo:
see #253, #260
set size policy on empty label to make it default to the amount of space available.

### DIFF
--- a/tilia/ui/windows/inspect.py
+++ b/tilia/ui/windows/inspect.py
@@ -50,7 +50,6 @@ class Inspect(QDockWidget):
     def __init__(self, main_window) -> None:
         super().__init__(main_window)
         self.setWindowTitle("Inspector")
-        self.setFixedWidth(300)
         self.setFeatures(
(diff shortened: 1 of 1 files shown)


FelipeDefensor (changes requested):
The code works but we still get the same problem as before, which is that the timeline elements may get moved when the selection changes, hurting control of the selection.


https://github.com/user-attachments/assets/38cef0ac-7dab-43f9-8d19-cc0091b6b01a

azfoo:
oops. There is an annoying bug where if the user sets the width by dragging the UI, the inspector stops flickering when the element selection changes. I don't know if it's fixable.

azfoo:
ps. got it.

FelipeDefensor (approved):
If there's an easy way to set it, I would like the initial width to be a little bigger, but don't bother too much with it.

FelipeDefensor:
> there is setting a minimum width but I don't know if that is what you want.

I was thinking about the width it first gets when opening. But if we can't set that then a slightly bigger minimum width could help.

PR #284, opened 2025-02-06
Fix crash when selecting file for import

FelipeDefensor:


FelipeDefensor:
We should be able to get this with automated tests. Will try to do it in the big PR for tests coming in the next few days.

Issue #272, opened 2025-01-31
Cannot re-open file, it has become unusable


johentsch:
Trying to open this file: https://github.com/MusicalForm/mozart_piano/blob/434bc3496e6aa52117288363da133faeefaabd48/TiLiA/Kristian%20Bezuidenhout%20-%20W.A.%20Mozart_%20Fantasie%20in%20d%20(KWV%20397).tla (the file is probably identical with the one in [this ZIP](https://github.com/user-attachments/files/18617238/Kristian.Bezuidenhout.-.W.A.Mozart_.Fantasie.in.d.KWV.397.zip) which also contains the audio), the error message "List index out of range" appears. When closing it, the app crashes with the report:

```python-traceback
During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "tilia\ui\actions.py", line 146, in callback
  File "tilia\requests\post.py", line 220, in post
  File "tilia\app.py", line 122, in on_open
  File "tilia\app.py", line 190, in on_restore_state
  File "tilia\app.py", line 202, in recover_to_state
  File "tilia\app.py", line 287, in on_clear
  File "tilia\timelines\collection\collection.py", line 331, in clear
  File "tilia\timelines\collection\collection.py", line 166, in delete_timeline
  File "tilia\timelines\base\timeline.py", line 203, in delete
  File "tilia\timelines\beat\timeline.py", line 125, in clear
  File "tilia\timelines\beat\timeline.py", line 73, in delete_component
  File "tilia\timelines\base\timeline.py", line 441, in delete_component
  File "tilia\requests\post.py", line 220, in post
  File "tilia\ui\timelines\collection\collection.py", line 276, in on_timeline_component_deleted
  File "tilia\ui\timelines\base\timeline.py", line 216, in on_timeline_component_deleted
KeyError: 3
```

In the last commit I had changed the measure numbers to reflect the repeats. I haven't checked until which point in the commit history the file was still working.

FelipeDefensor:
Fixed by 3019a4a0. Release available at: tilia-app.com/downloads.

PR #288, opened 2025-02-07
I'm done with parent and children attributes

FelipeDefensor:
The `parent` and `children` attributes of hierarchy components are completely determined by their `level`, `start` and `end` attributes. Therefore, storing them is redundant. Caching them would make sense if they needed to be retrieved frequently and the retrieval was expensive, but that is not the case. This PR introduces the `get_children` and `get_parent` methods, which are fast enough to keep a timeline with ~1500 components fairly responsive when grouping, merging, splitting, etc.

Additionally, making sure that those attributes are always updated has proved very hard. #272, #205 and an issue @johentsch has communicated to me directly were caused by that, and dozens of my personal files had to be manually adjusted because of dangling ids over the past years. This is made even worse by the fact that dangling ids may go undetected until the user tries to open a file, at which point she has lost access to it.

If that was not enough, getting rid of those attributes allows us to remove `TimelineComponent.SERIALIZABLE_BY_ID` and `TimelineComponent.SERIALIZABLE_BY_ID_LIST` and thus simplify significantly the serialization proccess. We will probably need attributes to reference timeline components in the future (e.g. to annotate variation relations), but I believe we should implement those from scratch to avoid the silent bugs we had up to now.


azfoo (approved):
hooray!!

PR #286, opened 2025-02-06
Towards less stressful releases (i.e. add a bunch of new tests)

FelipeDefensor:

PR #289, opened 2025-02-10
Even more tests

FelipeDefensor:
The bug fixed by e45771 could have been prevented if our tests covered `tilia.ui.dialogs.file.ask_for_file_to_open`. This PR accomplished that by refactoring the CSV import code to allow for more focused mocking, happening at a level lower than that of said function.

This depends on #286.


tilia/ui/ui_import.py
```diff
@@ -0,0 +1,141 @@
...
+from tilia.requests import get, Get
+from tilia.timelines.timeline_kinds import TimelineKind as TlKind
+from tilia.ui.dialogs.by_time_or_by_measure import ByTimeOrByMeasure
+from tilia.ui.timelines.collection.collection import TimelineUIs
+from tilia.parsers import get_import_function
+
+
+def on_import_from_csv(
+    timeline_uis: TimelineUIs, tlkind: TlKind
+) -> tuple[Literal["success", "failure", "cancelled"], list[str]]:
```
azfoo, line 14: this could probably be an enum

azfoo (approved):

tilia/ui/ui_import.py
FelipeDefensor: Agree. The rest of the import tests should also be refactored, we can make a general enum then. Would do it now but I don't want to dealy the release.

PR #290, opened 2025-02-10
Timelines scroll

### LINKED ISSUE #274: [Request] Score following with steady views
There is an alternative version of score-following which quite a few users might actually prefer. The current implementation leaves the play cursor in the middle of the window and shifts the score. The alternative would be to keep the view steady and shifting only when the cursor reaches the right end of the window, which is also what many score-following videos do. My suggestion would be to turn this into a Score-following setting with three possible values, something along the lines of `Off, Centered, Page-wise`.

### LINKED ISSUE #275: Auto-scroll not smooth with score timeline
When a file contains a score timeline and Auto-scroll is activated, the view jumps to the beginning of each note and makes it difficult to follow. For a demo: https://github.com/MusicalForm/mozart_piano/blob/main/TiLiA/Kristian%20Bezuidenhout%20-%20W.A.%20Mozart_%20Fantasie%20in%20d%20(KWV%20397)%20with%20score.tla

**Suggestion**

The score-following could simply follow the slider smoothly.

Loosely related: #267, #274

azfoo:
- implements scrolling timelines by page (closes #274)
- redo scroll (again, but hopefully for the last time) (closes #275 )


azfoo:
warning: running this version of code breaks all previous version's settings as auto-scroll becomes an enum.

FelipeDefensor (approved):
The new feature seems to work great and the commits were split in a way that was very easy to read. Good job!

- The were no tests for the new functionality. As it would be good to release this today with the rest of the approved PRs, I added them myself. I think we can do better on making adding tests a habit (me included).
- I also had to fix a failing test. Did you remembre to run them before making the PR? We should use GH actions to automate test runs on PR creation.
- I also added minor improvements to readability.

FelipeDefensor (changes requested):
I noticed two bugs after reviewing:
- Opening/creating a new file after opening another freezes the UI.
- Slider does not move if we open a file. Zooming fixes it.

FelipeDefensor:
Also, I get lots of this warning whenever I play or seek: "QtWarningMsg: QAbstractAnimation::pause: Cannot pause a stopped animation".

azfoo:
thanks! I completely forgot about tests, sorry.
I fixed the warning, but I'm not sure what you mean by the freezing. Could you show me?

azfoo:
If that's all, should we wrap this up and merge?

FelipeDefensor:
> Slider does not move if we open a file and play media. Zooming fixes it.

Is this fixed?

FelipeDefensor:
> seems to be on my end

Try this file, it doesn't work here. Tried on a couple others and it still did not work.

[pr-290-slider-not-moving.zip](https://github.com/user-attachments/files/18787324/pr-290-slider-not-moving.zip)

azfoo:
Right. I'm not sure we're seeing the same thing. I seem to have something wrong with `Player._engine_unload_media()` only when the first file has media played. Then when trying to load the second file, TiLiA stops responding at that command. But if the first file is not played, the second is able to load and play just fine. Something is definitely broken.

FelipeDefensor:
> Right. I'm not sure we're seeing the same thing. I seem to have something wrong with `Player._engine_unload_media()` only when the first file has media played. Then when trying to load the second file, TiLiA stops responding at that command. But if the first file is not played, the second is able to load and play just fine. Something is definitely broken.

Ok, so that's a very annoying bug that inconsistently pops up. I will try, again, to fix it. But I was talking about this other thing:

https://github.com/user-attachments/assets/64e0d95a-f917-4212-b922-977de89df710

azfoo:
ps. just to confirm, is smooth-scroll on in settings?

FelipeDefensor:
> ^so that's potentially a fix.

Well, that was fast. Thanks! Can you tell me what was wrong before? I'm just curious.

> And I have absolutely no clue why that is happening to your file. That is not something I can reproduce. One other annoying bug is the timelines getting rearranged when reopening a file. If the beat timeline gets created after score timeline, the beat tracking doesn't work.

Timeline creation order, as opposed to display order, is not designed to be consistent, as far as I remember. I don't think it is a good idea to rely on it.

azfoo:
> > ^so that's potentially a fix.
> 
> Well, that was fast. Thanks! Can you tell me what was wrong before? I'm just curious.

From what I can tell, QMediaPlayer probably runs on threads and calling multiple commands at once gets things confused (hence the multiple sleeps in this module). We should probably wait for the relevant changed signal to fire before sending a different command but that's a rewrite that could potentially wait. Deleting and recreating a new player is just a quick hack without identifying the real problem.

azfoo:
think I'm done!

FelipeDefensor (approved):

FelipeDefensor:
Why don't you try to trigger a build with these changes after merging?

1. Create a commit changing the version number in `setup.cfg` with the message "version: <version_number>"
2. Tag that commit with "v<version_number>"
3. Push the commit to dev
4. Push the tags with `git push --tags`
5. Check the Actions tab to see if the build was triggered

PR #293, opened 2025-02-17
Logging and Error Reports

azfoo:
Set up logging and error reporting with Sentry


azfoo:
ps. Check your email for the invite

tilia/ui/dialogs/crash.py
```diff
@@ -40,8 +55,78 @@ def _setup_widgets(self):
...
+    def __init__(self, parent) -> None:
+        super().__init__(parent=parent)
+        self.setWindowTitle("Contact Support")
+        self._setup_widgets()
+
+    def _setup_widgets(self):
+        self.setLayout(QFormLayout())
+
+        self.layout().addRow(
+            QLabel("Submit your contact details and we'll be in touch!")
```
FelipeDefensor, line 82: Given the user just experienced a crash, I would use a less cheerful message.

FelipeDefensor (changes requested):
This seems to work and I'm excited to see it in production! Also, thanks for cleaning up those stale files and functions. Some things to solve before merging:

- Add new dependencies (e.g. sentry-sdk) to requirements.txt and setup.cfg
- Tests for crash dialog and contact support dialog
- Sentry environment is alway set to production? 
image
- Get .tla dump: currently the app state is being printed as a line in the log. Could we get it in a separate file to make the log more readable and to facilitate reproducing bugs?
- Ensure performance is acceptable: I'm worried about the impact logging request might have on performance, as `post` and `get` are called all the time and it seems that each log triggers a write to a file. Let's see some profiling for those? I'm sure we can optimize them if needed.

tilia/ui/dialogs/crash.py
azfoo: Any suggestions?
Realistically, this is mostly to recover a lost file. I am thinking that this should also exist as a separate dialog that the user can call voluntarily to recover a file from a previous session if necessary. 
Sentry will send a report whenever an error occurs whether or not user data is set.

azfoo:
> * Sentry environment is alway set to production?

Is there an easy way to set an environment value in the gh workflow or some other place before publishing? Ideally, I would like to  turn Sentry off entirely for everything that isn't production (ie. dev, tests etc)

Will try to get the rest sorted.

FelipeDefensor:
> Is there an easy way to set an environment value in the gh workflow or some other place before publishing? Ideally, I would like to turn Sentry off entirely for everything that isn't production (ie. dev, tests etc)

Maybe we should use a .env file, beacuse that is also available at runtime in production and can be used in dev environments. But yes, we can set environment variables when building with GH actions: https://docs.github.com/pt/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables.

azfoo:
We have an .env file! I was wondering if I could let the logger read from there but we somehow manipulate that variable before publishing?

FelipeDefensor:
> We have an .env file! I was wondering if I could let the logger read from there but we somehow manipulate that variable before publishing? I just wasn't sure how we would set up that manipulation...
> 
> (Would really appreciate if you have the time to look into that -- would like this to be set up before the analysis-a-thon)

Sure! I think we can just do that using plain Python in `tilia.spec`. What do you want to set? Will look into that tomorrow.

azfoo:
Thanks! I'm thinking potentially two separate variables: 
- sentry-env: this sets a string/enum (perhaps `Production`, `Development`, `Testing`)
- sentry-dsn: this is the project to which logs are sent; this is currently set in init, but could be set to `None` if we don't want to log to sentry (like in dev/tests)

FelipeDefensor:
> Thanks! I'm thinking potentially two separate variables:
> 
>     * sentry-env: this sets a string/enum (perhaps `Production`, `Development`, `Testing`)
> 
>     * sentry-dsn: this is the project to which logs are sent; this is currently set in init, but could be set to `None` if we don't want to log to sentry (like in dev/tests)

See #294. I have created a general `ENVIRONMENT` environment variable. You can use that in a conditional to set the dsn or create a separate variable for it and change before building in `tilia.spec` (see example in the commits in #294). I think the conditional is simpler.

azfoo:
> * Ensure performance is acceptable

We are only logging posts (except those excluded in the .env file). Should we log gets also? I think most gets can also be inferred from the posts.

On loading a file with beat + score timelines (average of 5 tests each):
| Console + File + Sentry | File + Sentry | No Logging |
| --- | --- | --- |
| 10.155 | 9.4384 | 8.7076 |

I'd say logging is necessary if we want to have any kind of meaningful crash report. I have turned off console logging in production (which gives about 7.59% better time than with it). Not sure that makes a significant difference.

FelipeDefensor:
> > * Ensure performance is acceptable
> 
> We are only logging posts (except those excluded in the .env file). Should we log gets also? I think most gets can also be inferred from the posts.
> 
> On loading a file with beat + score timelines (average of 5 tests each):
> Console + File + Sentry 	File + Sentry 	No Logging
> 10.155 	9.4384 	8.7076
> 
> I'd say logging is necessary if we want to have any kind of meaningful crash report. I have turned off console logging in production (which gives about 7.59% better time than with it). Not sure that makes a significant difference.

That looks acceptable to, I would keep logging postst even in production. I agree logging is necessary for the report, but I'm not sure how useful simply echoing the requests will be, as the requests themselves were not designed with that in mind. But I think that is a good starting point.

About logging `get`, I think it is worth it iff you can easily add that to the profiling and see that the impact is of a similar degree.

azfoo:
Logging with everything on and gets is 12.4216s.

FelipeDefensor:
> Logging with everything on and gets is 12.4216s.

Tough call. I say we keep them out at first and wait for the first reports to see if adding them would help.

azfoo:
That is all for now. I'm not too sure how to do testing. What exactly should I be testing for?

FelipeDefensor:
> That is all for now. I'm not too sure how to do testing. What exactly should I be testing for?

Here is a impromptu list which we can work on together if you wish:

- Crash handler:
    - Is displayed when raising a random error
    - Opens the contact support dialog when clicking the appropriate button
This is a modal dialog, see `TESTING.md` for how to test those.
- Contact support dialog:
    - Sends an e-mail when confirmed
    - Stores credentials when Remember Me is checked
    - Does not store credentials when Remember Me is not checked

azfoo:
I have been trying to raise an exception unsuccessfully.  If you could help with that, I can sort the rest out?

tilia/ui/dialogs/crash.py
FelipeDefensor: And I like the idea of a recover dialog, I think it is what users expect instead of having to navigate to an autosave folder.

FelipeDefensor (approved):

FelipeDefensor:
Wait. I'm getting this error:

Traceback (most recent call last):
  File "C:\Users\Felipe\dev\tilia-desktop\tilia\main.py", line 1, in <module>
    from tilia.boot import boot
  File "C:\Users\Felipe\dev\tilia-desktop\tilia\boot.py", line 6, in <module>
    import dotenv
  File "C:\Users\Felipe\dev\tilia-desktop\.venv\Lib\site-packages\dotenv\__init__.py", line 3, in <module>
    from .main import (dotenv_values, find_dotenv, get_key, load_dotenv, set_key,
  File "C:\Users\Felipe\dev\tilia-desktop\.venv\Lib\site-packages\dotenv\main.py", line 2, in <module>
    import logging
  File "C:\Users\Felipe\dev\tilia-desktop\tilia\logging.py", line 3, in <module>
    import sentry_sdk
  File "C:\Users\Felipe\dev\tilia-desktop\.venv\Lib\site-packages\sentry_sdk\__init__.py", line 1, in <module>
    from sentry_sdk.scope import Scope
  File "C:\Users\Felipe\dev\tilia-desktop\.venv\Lib\site-packages\sentry_sdk\scope.py", line 12, in <module>
    from sentry_sdk.attachments import Attachment
  File "C:\Users\Felipe\dev\tilia-desktop\.venv\Lib\site-packages\sentry_sdk\attachments.py", line 4, in <module>
    from sentry_sdk.envelope import Item, PayloadRef
  File "C:\Users\Felipe\dev\tilia-desktop\.venv\Lib\site-packages\sentry_sdk\envelope.py", line 5, in <module>
    from sentry_sdk.session import Session
  File "C:\Users\Felipe\dev\tilia-desktop\.venv\Lib\site-packages\sentry_sdk\session.py", line 4, in <module>
    from sentry_sdk.utils import format_timestamp
  File "C:\Users\Felipe\dev\tilia-desktop\.venv\Lib\site-packages\sentry_sdk\utils.py", line 71, in <module>
    logger = logging.getLogger("sentry_sdk.errors")
             ^^^^^^^^^^^^^^^^^
AttributeError: partially initialized module 'logging' has no attribute 'getLogger' (most likely due to a circular import)

azfoo:
that's new! no I don't.

azfoo:
It might be because I am a terrible person who doesn't run things in venv

azfoo:
ps. does commenting out `import sentry_sdk` in line 3 of tilia.logging solve anything for you?

FelipeDefensor:
> ps. does commenting out `import sentry_sdk` in line 3 of tilia.logging solve anything for you?

It does not :(

azfoo:
running it in a venv still didn't bring up the issue...

FelipeDefensor:
> running it in a venv still didn't bring up the issue...

I tried a fresh venv here and it still happened. Can you try to build the app to see if it reproduces?

FelipeDefensor:
Here is an explanation of what is going on: https://stackoverflow.com/questions/77943889/how-to-defend-against-accidentally-shadowing-the-standard-library

This does not happen in the built app, as `pyinstaller` does some magic with `sys.path`. Running python with the `-P` flag fixes this, but I think it is safer to rename `tilia.logging` to `tilia.log` ou `tilia.logging_`.

azfoo:
Right. I can confirm that it builds. Does renaming the file work on your end? That would be an easy fix.

FelipeDefensor:
> Right. I can confirm that it builds. Does renaming the file work on your end? That would be an easy fix.

Done!

PR #296, opened 2025-02-22
Some bugs

azfoo:
6090862aed1d8a63d68f01d30c755bca696331f5 should really be tested, but from what I've tried so far, with documents stored in the same folder, it is possible to send folders including all necessary components for a tla file and reopen on the other end with the right links. (ie, if a .tla file is set up to have its pdf + audio/video in the same file in the original, zipping and sending the entire file should let you reopen the entire thing without needing to relocate files)


FelipeDefensor:
This is great. I am making a release right now without this just to ensure any early downloader from the AaT get the other commits. I will look into this later today or tomorrow. I would say we need tests before releasing, file opening has been one of the more problematic things over time. I can write them, though, no problem.

azfoo:
can we do a release without 6090862 in that case? 9c95b72 and 523546e are actually problematic when opening files that don't have the right links

FelipeDefensor:
> can we do a release without [6090862](https://github.com/TimeLineAnnotator/desktop/commit/6090862aed1d8a63d68f01d30c755bca696331f5) in that case? [9c95b72](https://github.com/TimeLineAnnotator/desktop/commit/9c95b72de3ae043a06693cd21b0c8596adce4672) and [523546e](https://github.com/TimeLineAnnotator/desktop/commit/523546e99c424c3521f503bd794fd04a6929a13d) are actually problematic when opening files that don't have the right links

Do you think it is fine to wait a couple hours? I plan to have the tests ready by then.

azfoo:
oops...
I don't know if these tests are sufficient

FelipeDefensor:
> oops... I don't know if these tests are sufficient

Oh, great! No worries, I had not started working on them. Will take a look and add anything that's missing.

FelipeDefensor:
> can we do a release without [6090862](https://github.com/TimeLineAnnotator/desktop/commit/6090862aed1d8a63d68f01d30c755bca696331f5) in that case? [9c95b72](https://github.com/TimeLineAnnotator/desktop/commit/9c95b72de3ae043a06693cd21b0c8596adce4672) and [523546e](https://github.com/TimeLineAnnotator/desktop/commit/523546e99c424c3521f503bd794fd04a6929a13d) are actually problematic when opening files that don't have the right links

The tests and the feature will need a bit of work still. Can you `git cherry-pick` those commits into `dev` directly and trigger a `0.5.9` build? I can release it right after.

PR #298, opened 2025-02-23
Score bugs

azfoo:


FelipeDefensor (changes requested):
The fixes seem to work. I added some tests. I requesting changes because there is at least one test that is hanging. I didn't debug, but I believe I got the same behavior in #297, so one of the commits I made there should fix it.

PR #303, opened 2025-03-02
Proper Ctrl+click selection

FelipeDefensor:


azfoo:
did you also want a review on this or is it still being worked on?

FelipeDefensor:
> did you also want a review on this or is it still being worked on?

Yes, this is ready for review.

FelipeDefensor:
Thanks for the review! Why the forced push?

tests/ui/timelines/test_timeline_ui.py
```diff
@@ -1,12 +1,55 @@
...
 from tilia.timelines.timeline_kinds import TimelineKind
 from tilia.ui.actions import TiliaAction
 
 
+class TestControlSelect:
+    # ideally we would test this on every kind of element
+    # but I didn't find an abstract way to click on all of them
+
+    def test_single(self, marker_ui):
+        for i in range(10):
```
azfoo, line 16: why 10?

tilia/ui/timelines/collection/collection.py
```diff
@@ -612,10 +612,11 @@ def on_selection_box_select_item(
...
             return
 
-        was_selected = timeline_ui.select_element_if_selectable(element, item)
+        if should_select := timeline_ui.should_select(element, item):
+            timeline_ui.select_element(element)
 
         # keep track of selection triggers under selection box
-        if was_selected:
+        if should_select:
             if element in self.sb_items_to_selected_items:
```
azfoo, line 620: could this not be squashed into one block?

azfoo (changes requested):

azfoo:
force-pushed to rebase onto latest dev (partially so tests will run)

tilia/ui/timelines/collection/collection.py
azfoo: or rather just return earlyier if should_select is False

tests/ui/timelines/interact.py
```diff
@@ -25,6 +25,8 @@ def click_timeline_ui_view(
     modifier = {
         None: Qt.KeyboardModifier.NoModifier,
         "shift": Qt.KeyboardModifier.ShiftModifier,
+        "control": Qt.KeyboardModifier.ControlModifier,
+        "ctrl": Qt.KeyboardModifier.ControlModifier,
```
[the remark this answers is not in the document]
FelipeDefensor: Just so we don't have to remember which one to use, as both are common ways to refer to the key.

tests/ui/timelines/test_timeline_ui.py
FelipeDefensor: Just an arbitrary n >= 2. Will add a comment to clarify.

FelipeDefensor:
@azfoo, not sure you got the review request, as this is still being shown as "changes requested".

tests/ui/timelines/interact.py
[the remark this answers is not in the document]
azfoo: I think we should stick to one.  It is just making sure that we get the right key in a dictionary when testing.

azfoo (changes requested):
Just that one comment, otherwise we're good to go.

azfoo (approved):

PR #304, opened 2025-03-05
Add utility function, make icecream available when testing and make image paths more robust 

FelipeDefensor:
These are some small changes I made while trying (and failing) to parse tremolos.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -29,6 +29,7 @@ testing =
     pytest>=8.0.0
     pytest-env>=1.1.5
     colorama>=0.4.6
+    icecream>=2.1.0
 ci-tests =
     coverage>=7.0.0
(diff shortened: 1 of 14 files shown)


setup.cfg
```diff
@@ -29,6 +29,7 @@ testing =
     pytest>=8.0.0
     pytest-env>=1.1.5
     colorama>=0.4.6
+    icecream>=3.0.0
```
azfoo, line 32: icecream is currently at 2.1.4.

azfoo (changes requested):

setup.cfg
FelipeDefensor: Not sure where I got that number from. Fixed!

azfoo (approved):

PR #305, opened 2025-03-05
Test on push

azfoo:
See ~[last run](https://github.com/azfoo/TiLiA/actions/runs/13680606388)~ [here](https://github.com/azfoo/TiLiA/actions/runs/13701637545) for results.

Some notes:
- ~Linux tests are not working~
- ~`setup.cfg` lists `python_requires = >=3.10` but we use `StrEnum` which is >=3.11~
- `audioop` is deprecated in 3.13 (this is unfortunately a [pydub](https://github.com/jiaaro/pydub/issues/760) thing, I'm not sure how or if we want to fix it. Tests for 3.13 are failing for this reason.)
- macOS/ubuntu and windows testing produces different errors


.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,68 @@
...
+      with:
+        path: ${{ matrix.path }}
+        key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
+        restore-keys: |
+         ${{ runner.os }}-pip-
+    - name: Install Python dependencies
+      run: |
+        # prerequisites
+        python -m pip install --upgrade pip wheel
+        python -m pip install coverage flake8 pytest-cov coverage-badge~=1.1.0
```
FelipeDefensor, line 50: Why do we need these? Could we add them to a section in `setup.cfg` instead of hard-coding?

azfoo:
> My push didn't trigger the test. Is this because we haven't merged yet? Is there any way to trigger it from the PR?

Because I set it to only run on push to dev or main. I don't think there is any way of triggering a run manually, but you can comment out lines 3 - 7 of `.github\workflows\run-tests.yml` and replace it with ` on:  [push]`., then commit and push the entire thing. I'll remove it later when we're done with testing.
(Although we could and perhaps should also talk about when this runs . Additionally, I would like it to not continue on failing so we remember to make the necessary changes before finalising things but that could become very annoying.)

FelipeDefensor:
> > My push didn't trigger the test. Is this because we haven't merged yet? Is there any way to trigger it from the PR?
> 
> Because I set it to only run on push to dev or main. I don't think there is any way of triggering a run manually, but you can comment out lines 3 - 7 of `.github\workflows\run-tests.yml` and replace it with ` on: [push]`., then commit and push the entire thing. I'll remove it later when we're done with testing. (Although we could and perhaps should also talk about when this runs . Additionally, I would like it to not continue on failing so we remember to make the necessary changes before finalising things but that could become very annoying.)

We can trigger a worflow manually, but it is kind of a hack, I had to do it when testing the automatic builds. In any case, I would like to see this also triggered on PRs. Is that possible?

azfoo:
mostly done, except these tests are still failing:

`tests/test_app.py::TestFileLoad::test_media_path_exists - assert 9.952 == 101
tests/ui/cli/test_load_media.py::test_load_with_forward_slashes - assert 100 ± 1.0e-04 == 9.952
tests/ui/timelines/beat/test_beat_timeline_ui.py::TestLoadFromFile::test_file_with_measures_to_force_display - AssertionError: assert ['1', '2', ''...'5', '6', ...] == ['1', '2', '3...'5', '6', ...]`

.github/workflows/run-tests.yml
```diff
@@ -5,6 +5,9 @@ on:
     branches:
       - main
       - dev
+  pull_request:
+    types: [opened, reopened]
+
```
FelipeDefensor, line 10: It would be nice to run when the PR is edited as well to check if the changes break something.

FelipeDefensor:
> * `tests/ui/timelines/beat/test_beat_timeline_ui.py::TestLoadFromFile::test_file_with_measures_to_force_display - AssertionError: assert ['1', '2', ''...'5', '6', ...] == ['1', '2', '3...'5', '6', ...]` I still have no clue what's wrong with this.

That *is* weird. Could you dig a bit more into this?

azfoo:
> > * `tests/ui/timelines/beat/test_beat_timeline_ui.py::TestLoadFromFile::test_file_with_measures_to_force_display - AssertionError: assert ['1', '2', ''...'5', '6', ...] == ['1', '2', '3...'5', '6', ...]` I still have no clue what's wrong with this.
> 
> That _is_ weird. Could you dig a bit more into this?

Tbh, I don't understand the original test. If we're setting `measures_to_force_display` and settings has s`display_measure_periodicity` at 4, shouldn't we be asserting that only ['1', '3', '5', '7'] are visible? (Even though that is not the logic of what is displayed at the moment.)

FelipeDefensor:
> > * `tests/test_app.py::TestFileLoad::test_media_path_exists - assert 9.952 == 101` - Media is loaded on macOS  & Linux, so media_duration is updated to the duration on file. Not sure how to resolve this.
> 
> What I think: there is a problem with the test, it sets the duration manually while it shouldn't. However, just correcting it doesn't work because loading a media is asynchronous. I'm trying to tackle async tests so we can fix this.

The issue was much simpler, after all. It had to do with the bad test and with "\" as separators in Windows. The load failed in Windows because of that, which made the test pass when it should fail. We didn't catch this because we are not handling errors messages by `QMediaPlayer`. I will add a task for that in PM.

FelipeDefensor:
> Tbh, I don't understand the original test. If we're setting `measures_to_force_display` and settings has `display_measure_periodicity` at 4, shouldn't we be asserting that only ['1', '3', '5', '7'] are visible? (Even though that is not the logic of what is displayed at the moment.)

This comment made eveything clear. The test only passes on my machine becaus I had the display measure periodicity set to 2 on earlier runs. 1, 3, 5, 7 are the odd measure that wouldn't be displayed if not forced.

FelipeDefensor (changes requested):
Please see if the tests are passing now. Apart from that, see if we can trigger the test runs whenever there is a push to PR, not just when it is first opened.

FelipeDefensor (approved):
Hooray!

PR #308, opened 2025-03-12
Improve exported JSON

FelipeDefensor:
This makes files exported with the export feature more aligned with what JSON users would expect. We were using lists instead of objects to store components to reduce duplication. However, compression can take care of the key duplication in a size-sensitive context. Using object has the benefit of making the output format less complex, less surprising and easier to document.

Also added some tests for the export function and some fixes to test convenience functions.

This is made in preparation to documentation on the export feature.

### DIFF
--- a/tests/mock.py
+++ b/tests/mock.py
@@ -2,7 +2,7 @@ from contextlib import contextmanager
 from typing import Any, Sequence
 from unittest.mock import patch, Mock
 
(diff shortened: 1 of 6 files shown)


FelipeDefensor:
@azfoo, checks are now passing. Review when you can.

azfoo:
force-pushed for squashes, removing an outdated function and fixing a naming typo.

azfoo (approved):

PR #309, opened 2025-03-12
Joss submission

azfoo:
@FelipeDefensor: not sure if we want to leave this branch here for submission or merge it


azfoo:
> Thanks! We should delete the corresponding files from the other repo, right?

Will do that.

Actually, should this live on main or dev?

FelipeDefensor:
> > Thanks! We should delete the corresponding files from the other repo, right?
> 
> Will do that.
> 
> Actually, should this live on main or dev?

Let's put it on dev, not sure what kind of changes might be requested during the review.

Issue #312, opened 2025-03-13
Clefs are parsed to wrong staves


azfoo:
In preparing an image for the JOSS paper, I discovered some bugs. Most have been fixed here: #317.

The last one which has yet to be solved is reopening files with score_tls:
- clefs are swapped
- notes are also shifted

![Image](https://github.com/user-attachments/assets/9e4a589a-7c78-4e6e-9a44-dbb84d731920)

FelipeDefensor:
Thanks for spotting this! What do you mean by "shifted"? Please send the tla and the mxl so I can try to reproduce.

azfoo:
[Files here](https://github.com/user-attachments/files/19228580/k310.no1.zip)

What I mean is that they show up fine on the import but get swapped when reopening the file.

azfoo:
ps. I haven't fixed the clefs but have accidentally discovered why the notes are a third higher.

FelipeDefensor:
Fixed by #319

PR #317, opened 2025-03-14
More score bugs

azfoo:
Some bugs mentioned in #312, including:
- incorrect parsing of clefs
- incorrect clef/key signature/time signature when not importing the first measure
- overlapping time signature in block display


FelipeDefensor (approved):
Glad you found the third higher bug. I wasn't able to, even after several hours.

I'm approving this as it has important fixes, but please write tests for the buggy behaviors after meging.

PR #319, opened 2025-03-17
Even more score bugs

### LINKED ISSUE #312: Clefs are parsed to wrong staves

FelipeDefensor:
In combination with #317, closes #312.


azfoo (changes requested):
72d2fb5 has revealed a few problems with the way we cache notes, particularly because the order in which elements are imported is not fixed. Will work on this.

azfoo:
- force-pushed to start writing tests on issues found in #312 
- lists (72d2fb5) make missing staff numbers fall through the cracks and is not a good idea

azfoo:
> > ```
> > * lists ([72d2fb5](https://github.com/TimeLineAnnotator/desktop/commit/72d2fb5138a5a1ef5e27a097e7d71bb431711896)) make missing staff numbers fall through the cracks and is not a good idea
> > ```
> 
> Yes, and I did not think about that. But are missing staff numbers those any more possible than duplicated staff numbers, which would be overwritten if using dictionaries?

Will check for that and raise an error.

azfoo:
> > ```
> > * lists ([72d2fb5](https://github.com/TimeLineAnnotator/desktop/commit/72d2fb5138a5a1ef5e27a097e7d71bb431711896)) make missing staff numbers fall through the cracks and is not a good idea
> > ```
> 
> Yes, and I did not think about that. But are missing staff numbers those any more possible than duplicated staff numbers, which would be overwritten if using dictionaries?

I mean missing as in imported in the wrong order. 
For example, attempting to import a score component before the staff is imported, causing a reference in the list to an element that is not present - the list of extreme notes could potentially refer to an element in a completely different staff once the ids are corrected. 
Still, I could try to validate the staff numbers.

azfoo (approved):

azfoo:
@FelipeDefensor: could you review this when you have the chance?

tests/ui/timelines/score/test_score_timeline_ui.py
```diff
@@ -43,3 +47,56 @@ def test_create_key_signature(score_tlui, fifths):
...
+        ):
+            components = score.timeline.get_components_by_attr("KIND", cmp_kind)
+            staff_no_to_y = {
+                cmp.staff_index: score.get_element(cmp.id).body.y()
+                for cmp in components
+            }
+            sorted_y = [
+                k for k, _ in sorted(staff_no_to_y.items(), key=lambda item: item[1])
+            ]
+            assert len(sorted_y) == 3
```
FelipeDefensor, line 71: Why this hardcoded 3 here? Are we expecting a specific number of components in the timeline? If so, let's explicitly acknowledge that.

FelipeDefensor (commented):
The implementations seem to work. We need tests for importing with empty staves and with invalid staff numbers.

FelipeDefensor:
@azfoo, I can't review beacus I'm the author of this PR, so I left my review comments above.

azfoo:
> The implementations seem to work. We need tests for importing with empty staves and with invalid staff numbers.

Done now. Anything else?

tilia/ui/timelines/score/timeline.py
```diff
@@ -435,12 +442,13 @@ def on_score_timeline_components_deserialized(self, id: int):
...
         self.staff_numbers = sorted(self.staff_numbers)
-        if [*set(self.staff_numbers)] != self.staff_numbers or sorted(
-            self.staff_extreme_notes
-        ) != self.staff_numbers:
-            tilia.errors.display(tilia.errors.SCORE_STAFF_ID_ERROR, self.staff_numbers)
+        return (
+            [*set(self.staff_numbers)] == self.staff_numbers
+            and set(self.staff_extreme_notes).issubset(self.staff_numbers)
+            and set(self.staffs_with_symbol).issubset(self.staff_numbers)
+        )
```
FelipeDefensor: 1. Didn't know you could unpack sets. Python is always more magical than you think.
2. Good refactor, the responsabilities are better distributed and the code is easier to read.

FelipeDefensor:
By me, this is good to merge.

PR #323, opened 2025-03-29
Export timelines to image

azfoo:


tilia/ui/qtui.py
```diff
@@ -111,6 +119,30 @@ def closeEvent(self, event):
...
+        widget: QGraphicsScene = self.centralWidget().scene()
+        success, result = ResizeRect.new_size(
+            widget.sceneRect().width(), widget.sceneRect().height()
+        )
+        if not success:
+            return
+
+        if result[0] != widget.sceneRect().width():
+            zoom_level = result[0] / widget.sceneRect().width()
+            post(Post.VIEW_ZOOM_IN, zoom_level)
```
FelipeDefensor, line 133: What if the user chooses a size that is bigger than the current? Does `Post.VIEW_ZOOM_*IN*` still work?

FelipeDefensor (changes requested):
This needs a bit of work.
- I'm bothered that there is no default name for the exported image in the Windows file dialog (which is also the case when saving .tlas);
- The default name filter should be PNG or JPEG;
- The "resize timelines" dialog UI can be improved:
  - "Resize timelines" as a title makes me think that the timeline's UI will be resized, not that I'm setting the exported image's size.
  - Dialog icon is not TiLiA icon.
  - Displaying "Original", "New height" and "New Width" seems cluttered to me. I would simply set original as the spinbox's initial value and call the fields 'Widht" and "Heighth".
- We also need tests.

FelipeDefensor:
Issues aside, this seems to work, so we can use it to get images for the AaT.

tilia/ui/qtui.py
azfoo: Yes it does. It's just maths magic!

FelipeDefensor:
Oh, just remembered we need to update the docs on the website after releasing this.

azfoo:
> Issues aside, this seems to work, so we can use it to get images for the AaT. I wasn't able to get anything too useful, though, as we have to zoom in pretty far to make components not overlap.

I should have updated the other repo with images already. 
I know they appear ridiculously long but I think of it more as a substitute for people who don't want to download the app and would like to still have a copy of the results. 
We could probably add a resize/crop function at some other point (if exporting to images proves useful) that would let users choose which section of the analysis they want to export. (That's just editing `source` of `scene.render()`.)

FelipeDefensor:
Much better! But reading the tests made me realize that setting a timeline as not exportable to JSON also makes it not exportable as image, which is not what we want. Maybe set different flags for each case?

azfoo:
> setting a timeline as not exportable to JSON also makes it not exportable as image

Not true. Every object on TimelinUIsScene is exported. The only way to not export something is for the object to not be visible at the time of exporting. At the moment, the conventional score is not being exported because that lives in a different window. (That is something I hope to implement at some point. Deciding how to line up the two images or to export them separately is a different question.)

FelipeDefensor:
> > setting a timeline as not exportable to JSON also makes it not exportable as image
> 
> Not true. Every object on TimelineUIsScene is exported. The only way to not export something is for the object to not be visible at the time of exporting. At the moment, the conventional score is not being exported because that lives in a different window. (That is something I hope to implement at some point. Deciding how to line up the two images or to export them separately is a different question.)
> 
> See the images in the other repo: slider timeline (which has the NOT_EXPORTABLE flag) is shown.

You're absolutely right, sorry. Here's a reminder for me not to review other's work while tired in the weekend 🙃.

FelipeDefensor (approved):

PR #324, opened 2025-03-29
Alt-click to select without seeking

### LINKED ISSUE #204: Selecting a component should probably not move the cursor/slider
Selecting the exact position on the slider is one of the central tasks when annotating. Having spent some time to stop playback at a precise position, it causes great pain to lose it again, e.g. because one forgot to add the child to be split. Currently, there is no way (that I know of) to select a node without losing the cursor position. In my view, the value of a precisely selected position would warrant to make moving it to a component's beginning optional (e.g. through `Alt + click` which currently duplicates the functionality of `Ctrl + click`).

FelipeDefensor:
Depends on #303.

Closes #204.


azfoo:
please rebase onto the latest dev (or #303 since that is already updated).
also, I'm not sure what you expect alt+ ctrl + click to do. I thought it would let you select multiple without ever moving the playback line but it doesn't let you select multiple and it moves the playback line :(

FelipeDefensor:
> I'm not sure what you expect alt+ ctrl + click to do. I thought it would let you select multiple without ever moving the playback line but it doesn't let you select multiple and it moves the playback line :(

It should definitely do what you expect, but I forgot to implement that. Nice catch!

FelipeDefensor:
@azfoo: fixed it!

FelipeDefensor:
Also, let's release 0.5.11 after this one?

azfoo (changes requested):
Are we expecting there to be a need for alt-unselect as well?

ps. sorry for the failing tests, see commit message.

azfoo (approved):

PR #325, opened 2025-03-31
Automatically update media metadata title on save if not previously set

azfoo:


FelipeDefensor (approved):
This is a nice touch.

PR #344, opened 2025-06-20
Assorted commits preparing for the bpsd-to-tilia repo

FelipeDefensor:
Some features/fixes to the CLI were necessary for converting the BPSD dataset. This needs to be merged on checked out in order to test the code there.


FelipeDefensor:
This will be split into two more cohesive commits.

PR #368, opened 2025-09-02
Fix using media relative path when .tla is moved (#364)

### LINKED ISSUE #364: Cross-system compatibility regarding filepaths (+ reducing number of file changes)
**Scenario** 

Using TiLiA version 0.5.14 on Windows to review form analyses created on a different Windows computer using version 0.5.12. The files in question point to local MP3 recordings. Changes to the TiLiA file are being tracked using git.

**Desired behaviour**

- Opening the TiLiA file works without error.
- Since on my computer, the MP3 is in the same location, relative to the TLA file, as on the other computer, I can click play and it starts playing.
- When I store the file, no change is introduced in the TiLiA file because nothing has changed.

**Actual behaviour**

- In some cases, opening a file results in a "Media load error" (see hidden screenshot below) after which I get a "Invalid media path" prompt asking me whether or not I would like to load another media file. When I click yes, I get to select the correct file location on my system.
- In other cases, when the automatic inference of the media file's location works (it is mysterious when it does and when it doesn't), the play button, upon clicking, toggles to the "pause" symbol but playback does not start. Playback does not work after re-opening the file either.
- When storing the file, not only the filepaths change (because they are absolute paths), but also the format of the media path:
  Image
  This, however, is inconsistent, in other cases, the *nix format with forward slashes is preserved:
  Image
- The IDs of all components change, leading to a large number of redundant changes to the file when, actually, nothing changed (except the TiLiA version). Not to mention changing IDs defying the very idea of IDs.

<details><summary>Media load error</summary>
<p>

Image

</p>
</details> 

**Proposed solution**

The commonplace patterns here (apart from using a zip-based file format that includes media files) consist in 

- always ever storing *nix-style paths, never Windows-style with backslashes;
- relying on relative paths exclusively, which (in many cases) also resolves the security issue of revealing home-directory locations and user names;
  - in order to avoid using `../`, TiLiA could detect the common ancestor in the filepaths of the TLA and the media file and store them relative to that ancestor.

**Example:**

- TLA file: `C:\Users\hentschel\git\annotation_pilot\TiLiA\Haydn_Sonatas\haydn-sonata_XVI_14_D_2_menuet.tla`
- MP3 file: `C:\Users\hentschel\git\annotation_pilot\recordings\haydn_sonatas\haydn-sonata_XVI_14_D_2_menuet_mod.mp3`
- Common ancestor: `C:\Users\hentschel\git\annotation_pilot\`
- Stored paths:
  - `TiLiA/Haydn_Sonatas/haydn-sonata_XVI_14_D_2_menuet.tla`
  - `recordings/haydn_sonatas/haydn-sonata_XVI_14_D_2_menuet_mod.mp3`

FelipeDefensor:
This also improves the workaround to avoid freezes when loading a new media file or unloading the current one.

### DIFF
--- a/tests/file/test_file_manager.py
+++ b/tests/file/test_file_manager.py
@@ -1,6 +1,9 @@
 import json
+import sys
 from pathlib import WindowsPath, Path
 from unittest.mock import mock_open
+
+import pytest
 
 from tests.constants import EXAMPLE_MEDIA_PATH
 from tests.mock import PatchPost, Serve, patch_file_dialog, patch_yes_or_no_dialog
@@ -138,6 +141,9 @@ class TestFileManager:
 
             post_mock.assert_called()
 
+    @pytest.mark.skipif(
+        not sys.platform.startswith("win"), reason="Windows specific test"
+    )
     def test_is_saved_with_posix_paths(self, qtui, tilia, tmp_path, user_actions):
         file_path = tmp_path / "test.tla"
         with patch_file_dialog(True, [str(EXAMPLE_MEDIA_PATH)]):
--- a/tests/file/test_file_manager.py
+++ b/tests/file/test_file_manager.py
@@ -1,5 +1,8 @@
 import json
+from pathlib import WindowsPath, Path
 from unittest.mock import mock_open
+
+from tests.constants import EXAMPLE_MEDIA_PATH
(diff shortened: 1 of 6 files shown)


FelipeDefensor:
Closes #364.

PR #375, opened 2025-11-26
Commands API

### LINKED ISSUE #395: Existing IDs are not being considered when choosing a new ID for a unit
b13eb27d makes IDs persist when saving/opening files and `App.get_id()` generates monotonically increasing integet IDs. The ID generator is reset to 0 when opening a file, so there is a very real chance that an ID collision will occur, especially if we are opening a file that already has small integers as IDs (as most files probably will).

To reproduce:
- Create new file
- Create a marker timeline
- Save file
- Open file
- Create another marker timeline
- Save file

Both marker timelines should have an ID of 1 (!), which means they will overwrite each other when opening the file in another session.

FelipeDefensor:
Introduces the commands API, a much simpler way to handle user interactions. Commands make all the following obsolete:

- All timeline and timeline component `RequestHandlers`
- `TimelineComponentSelector`
-  Several `TimelineSelectors`
- `TiliaActions`
- `user_action` fixture
- `TimelineUi.on_timeline_request`, `TimelineUI.on_timeline_component_request` (substituted by `TimelineUIs.on_timeline_command`)
- `TimelineUIs.pre_process_timeline_request()`
- `tilia\ui\timelines\collection\requests\args.py`
- Several `Post` and `Get` members
- Some `TimelineSelector` members

The core functionality is given by `command.py`, along with fuller documentation.

Usage of commands:
```
    # Register a new command
    commands.register(
        'example.command',
        callback_function,
        text='Menu Text',  # Optional: for display in Qt interface
        shortcut='Ctrl+E',  # Optional: keyboard shortcut
        icon='example_icon'   # Optional: icon name in IMG_DIR (without extension)
    )

    # Execute a command
    commands.execute('example.command', arg1, arg2, kwarg1=value1)
```

Examples of commands:
- File commands:        'file.open', 'file.save', 'file.export.img'
- Timeline commands:    'timeline.delete', 'timeline.component.copy'
- View commands:        'view.zoom.in', 'view.zoom.out'

Note that we can pass arguments to commands directly, so we don't need (horrible) things like `TimelineSelector.FROM_MANAGE_TIMELINES_TO_PERMUTE`.

Commands also set the stage for plugins, as they provide a simpler interface to modify application state programatically.

This PR also contains small changes that were covenient preliminaries for or complements to the implementation:
	- Bug fixes (3be8cee41664cc6157d9bb775ca0bf9f495fa942, 24576bddb619ab9e8c9ca3559988aa0e85621846, 6964fdd66faa38910addf783eac7710ffc37db30)
	- Additional testing (467cb5ed048ae5443218caa366ec066164c6980b, f07fd4da9304bac1695986d4ceeabce24e5edf3b)
	- Convenient refactors (b8d15ef49cfcb0cb904de54c0c9db368a0d62dfd,  914a0bd3a1cc925d644abe6540fca32f730cfb84, 974e71508f46ed562fc854746f0c9b68c6f1bc6a, 769e30ba35b73b763eb91cb08de12c2c00a0d462, f13fa08eb1255014fdb8033398c0b7213fc78287)
	- Documentation (45dd1dadace511835b260b1f694094e5e95d83b4)


FelipeDefensor:
This shaves of 600 lines, while still adding functionality and fixing a bunch of bugs. Feels good :)
grafik

FelipeDefensor:
@azfoo, let me know if you want to review this (partly or fully).

tests/ui/timelines/test_timeline_ui.py
```diff
@@ -276,3 +277,15 @@ def test_set_empty_string(self, tls, tluis, user_actions):
...
+def test_set_is_visible(tls, marker_tlui):
+    with undoable():
+        commands.execute("timeline.set_is_visible", marker_tlui, False)
+
+    assert marker_tlui.view.isVisible() is False
+
+    with undoable():
+        commands.execute("timeline.set_is_visible", marker_tlui, True)
+
+    assert marker_tlui.view.isVisible() is True
```
azfoo: assert not isVisible() / assert isVisible() ?

FelipeDefensor:
@azfoo, have you run into similar CI failures? Would they be fixed by the new build system?

azfoo:
A locally run Linux pytest fails on the same test. We might have to look into why.

tilia/timelines/base/timeline.py
```diff
@@ -33,11 +36,27 @@
...
 class Timeline(ABC, Generic[TC]):
     SERIALIZABLE = ["name", "height", "is_visible", "ordinal"]
     NOT_EXPORTABLE_ATTRS = []
     KIND: TimelineKind | None = None
-    FLAGS = []
+    FLAGS = [
+        TimelineFlag.HAS_COMPONENTS,
+        TimelineFlag.COMPONENTS_COPYABLE,
+        TimelineFlag.COMPONENTS_IMPORTABLE,
+    ]
```
azfoo: Could this be made into something more generic so not every timeline rewrites this?

tilia/timelines/base/timeline.py
```diff
@@ -110,6 +132,31 @@ def components(self):
     def default_height(self):
         return None
 
+    @classmethod
+    def subclasses(cls):
+        cls.ensure_subclasses_are_available()
+        return cls.__subclasses__()
```
azfoo, line 136: Can this be refactored so that `ensure_subclasses_are_available()` is only instantiated once and not every time the method is called?

tilia/timelines/base/validators.py
```diff
@@ -17,7 +17,7 @@ def validate_bool(value):
 
 
 def validate_color(value):
-    if value is None:
+    if value is None or value == "":
```
azfoo, line 20: is `if not value` not sufficient?

tilia/timelines/collection/collection.py
```diff
@@ -299,12 +299,20 @@ def crop_timeline_components(self, new_length: float) -> None:
...
+        else:
+            return None
 
     def get_metric_position(self, time: float) -> MetricPosition | None:
         if not self.get_timelines_by_attr("KIND", TimelineKind.BEAT_TIMELINE):
             return None
         tl = self.get_beat_timeline_for_measure_calculation()
+        if not tl:
+            return None
         beats = tl.components
```
azfoo, line 330: Will this if clause not already be caught by `get_beat_timeline_for_measure_calculation()`?

tilia/timelines/timeline_kinds.py
```diff
@@ -40,23 +22,19 @@ def get_timeline_kind_from_string(string):
...
+
+
+def get_timeline_frontend_name(kind: TimelineKind) -> str:
+    name = get_timeline_name(kind)
+    if kind == TimelineKind.PDF_TIMELINE:
+        name = name.upper()
+
+    return name
+
+
```
azfoo, line 36: Where is this method used?

azfoo:
Just to clarify, why refactor some requests into commands but not the others?

tilia/ui/timelines/base/context_menus.py
azfoo, line 1: After adding a new timeline, additional move up/down options do not appear
e.g. create a new file, add marker tl1, then marker tl2. (so from top to bottom: slider, tl1, tl2). right-clicking on tl1 should have options for both up and down, but only up is shown. however, once moved up, the menu options are correct.

p.s. should slider tl also have a context menu with up/down options?

FelipeDefensor:
> Just to clarify, why refactor some requests into commands but not the others?

The main goal here was to get rid of the convoluted logic around `TimelineUI.on_timeline_request` and `TimelineUI.on_timeline_component_request`. In the long term, the idea is to use the commands API for any action that might be performed by an user, and that could in later be used by plugin developers. Our "internal" requests should still use the `post\get` API, although at this point there still might be some things that are user actions that still use it.

tilia/ui/timelines/base/timeline.py
```diff
@@ -84,9 +108,25 @@ def __bool__(self):
...
+    def ensure_subclasses_are_available(cls):
+        timelines_dir = Path(os.path.dirname(__file__)).parent
+        packages = [
+            "tilia.ui.timelines." + d.name
+            for d in timelines_dir.iterdir()
+            if d.is_dir() and d.name not in ["base", "__pycache__"]
+        ]
+        for pkg in packages:
+            importlib.import_module(pkg)
+
```
azfoo, line 129: as above in timelines/base/timeline

azfoo:
In that case, I am wondering whether it would be easier to document if all of these "external" actions were centralised in one place, like in a JSON file (see: https://github.com/ArjanCodes/examples/tree/main/2026/spec for example). The current naming system of  `timeline.kind.action` also lends itself very nicely to become some kind of dict structure.

tilia/ui/timelines/collection/collection.py
```diff
@@ -116,6 +126,274 @@ def _setup_widgets(self, main_window: QMainWindow):
...
+        try:
+            for tlui in timeline_uis:
+                if method_name:
+                    # Use type(tlui) to get an unbound method, since we will pass tlui explicitly.
+                    callback = getattr(type(tlui), method_name)
+                result.append(callback(tlui, *args, **kwargs))
+        except Exception:
+            tilia.errors.display(
+                tilia.errors.COMMAND_FAILED,
+                f"Timeline command callback '{callback.__name__}'",
```
azfoo: `callback` might not have `__name__` if it turns out to not be a method

tilia/ui/timelines/collection/collection.py
```diff
@@ -116,6 +126,274 @@ def _setup_widgets(self, main_window: QMainWindow):
...
+
+        success = self.validate_command_return_value(
+            method_name or callback.__name__, result
+        )
+
+        if success:
+            post(
+                Post.APP_STATE_RECORD,
+                f"timeline command: {method_name or callback.__name__}",
+            )
```
azfoo: if `validate_command_return_value` only checks that the returned result is a list of bools / a bool and not if the command is actually successful, what happens after a failed command (result has a false) when the user tries to undo?

tilia/ui/timelines/collection/collection.py
```diff
@@ -116,6 +126,274 @@ def _setup_widgets(self, main_window: QMainWindow):
...
+    def on_timeline_command(
+        self,
+        kind: TimelineKind | list[TimelineKind],
+        callback: Callable | str,
+        selector: TimelineSelector,
+        *args,
+        **kwargs,
+    ):
+        """
+        To be used as a wrapper for a timeline command.
```
azfoo: but it isn't a wrapper?

FelipeDefensor:
Serious bug:

> In that case, I am wondering whether it would be easier to document if all of these "external" actions were centralised in one place, like in a JSON file (see: https://github.com/ArjanCodes/examples/tree/main/2026/spec for example). The current naming system of `timeline.kind.action` also lends itself very nicely to become some kind of dict structure.

This would have a lot of advantages, such as making it way easier to find the action definitions. We would probably have to add quite a bit of logic to be able to specify callbacks and validate callbacks from an external file, right?

PS.: ArjanCodes fan here :)

FelipeDefensor:
Bug: timeline toolbars are sometimes not being displayed for all timeline types in a file when opening a previously saved file. Here is a file where that happens: [example.tla.json](https://github.com/user-attachments/files/25867634/example.tla.json)

**This must be fixed before releasing.**

Edit: I was not able to reproduce this, I am not sure the changes here are the cause for the behaviour anymore.

Edit: I was able to reproduce with this file: [harmony-tl-empty.tla.json](https://github.com/user-attachments/files/25876697/harmony-tl-empty.tla.json)
I probably saved the other file while using this branch, which fixed it somehow.

tilia/ui/commands.py
```diff
@@ -0,0 +1,146 @@
...
+    if os.environ.get("ENVIRONMENT") != "prod":
+        return _execute_dev(command_name, *args, **kwargs)
+    else:
+        return _execute_prod(command_name, *args, **kwargs)
+
+
+def _execute_dev(command_name: str, *args, **kwargs):
+    if command_name not in _name_to_callback:
+        for key in _name_to_callback:
+            print(key)
```
azfoo: can we not use print but instead attach it to the error message?

tests/parsers/csv/test_markers_from_csv.py
```diff
@@ -118,39 +113,41 @@ def test_markers_by_time_from_csv_fails_if_no_time_column(marker_tlui):
...
+    marker_tlui, tilia_errors
+):
     data = "time\nnonsense"
 
-    status, errors = patch_import("time", marker_tlui.timeline, data)
+    success = patch_import("time", marker_tlui.timeline, data)
 
-    assert status == "success"
-    assert_in_errors("nonsense", errors)
+    assert success is True
```
azfoo: assert success?

azfoo (changes requested):

azfoo:
> This would have a lot of advantages, such as making it way easier to find the action definitions. We would probably have to add quite a bit of logic to be able to specify callbacks and validate callbacks from an external file, right?

Happy to help if needed. Definitely see it as something that is potentially quite nice to have.

tilia/app.py
```diff
@@ -275,11 +275,30 @@ def on_record_state(self, action, no_repeat=False, repeat_identifier=""):
...
+        component_ids = set()
+        for component_list in component_lists:
+            component_ids = component_ids.union({c.id for c in component_list})
+        existing_ids = timeline_ids.union(component_ids)
+
+        next_id = next(self._id_counter)
+        # Increments counter until a unique id is found.
+        # Could easily be optimised.
+        while str(next_id) in existing_ids:
+            next_id = next(self._id_counter)
```
azfoo: How does this perform when creating a new element after loading a very large file?
I was wondering about setting the start id (`itertools.count(start)`) after getting the largest id from after the file loads. That would save having to loop through everything each time you get an id.
id will keep growing but that might be helpful when comparing across versions.
Also of potential interest is a way to validate ids when loading a file.

FelipeDefensor: The performance is acceptable. In a file with thousands of hierarchies it just adds a ~1.5 second overhead to the ~12 seconds it was taking to open. Starting from the largest id like you suggested is more elegant and performant, though. It is also easy to implement. Will do that.

FelipeDefensor: Validating IDs could be good, yes. I am considering implementing some major changes to IDs for v0.7, so let's leave this for then, ok?

FelipeDefensor:
> > This would have a lot of advantages, such as making it way easier to find the action definitions. We would probably have to add quite a bit of logic to be able to specify callbacks and validate callbacks from an external file, right?
> 
> Happy to help if needed. Definitely see it as something that is potentially quite nice to have.

How about we first convert everything to commands as they are (see, for instance, #387) and then consider this more thorougly?

tilia/ui/commands.py
FelipeDefensor: This has been improved in another branch. I'll find the commit and move it here.

tilia/timelines/collection/collection.py
FelipeDefensor: I guess this is supposed to be merely defensive, but I may have got a crash on the `beats = tl.components` line the other day. Should have documented it.

tilia/timelines/base/validators.py
FelipeDefensor: Not sure if I did this intentionally, but I guess it defends against other falsey values, which we don't want to see there. Are you against leaving it there?

azfoo: more so that the acceptable types of QColor inits are known. a try/catch might work better because we could potentially crash in the next part.

tilia/ui/commands.py
FelipeDefensor: Fixed by 23c2eda5

tests/parsers/csv/test_markers_from_csv.py
FelipeDefensor: Fixed in df61c274

tilia/ui/timelines/collection/collection.py
FelipeDefensor: Fixed in eff94b91

tilia/app.py
FelipeDefensor: Fixed in 26ebdf7e

tilia/ui/timelines/collection/collection.py
FelipeDefensor: Fixed in 0793fd58

tilia/ui/timelines/collection/collection.py
FelipeDefensor: This was completely wrong. Thanks. Fixed by a9654aaf.

FelipeDefensor:
> Bug: timeline toolbars are sometimes not being displayed for all timeline types in a file when opening a previously saved file. Here is a file where that happens: [example.tla.json](https://github.com/user-attachments/files/25867634/example.tla.json)

> Edit: I was able to reproduce with this file: [harmony-tl-empty.tla.json](https://github.com/user-attachments/files/25876697/harmony-tl-empty.tla.json) I probably saved the other file while using this branch, which fixed it somehow.

I wasn't able to reproduce it anymore with that last file. This has happened again with another file, though. I am worried, but I don't know how we should proceed if we are not able to reproduce.

tilia/timelines/base/timeline.py
FelipeDefensor: I didn't understand the comment. Could you be more specific. I am not entirely satisfied with the solution I came up with. A clear improvement would be to use [`enum.Flag`](https://docs.python.org/3/library/enum.html#enum.Flag) instead of a list.

azfoo: Every timeline has rewritten this variable. I would prefer it if it were left as an empty list, or as a list of values that is shared by a number of them (probably copyable and importable).

PS. Is `TimelineFlag.HAS_COMPONENTS` used for anything?

tilia/timelines/base/timeline.py
FelipeDefensor: This was only called when a new timeline was instantiated, so the performance impact is negligible. Even then, it was an easy fix in 09d3d764.

tilia/timelines/base/timeline.py
FelipeDefensor: Got it. Makes sense. Fixed in 0b2fca43.

PS. Fixed in 3251835b.

tilia/timelines/base/validators.py
FelipeDefensor: How about what I did in 72bcd074?

tilia/timelines/collection/collection.py
FelipeDefensor: I am leaving this just in case, ok?

FelipeDefensor:
> I wasn't able to reproduce it anymore with that last file. This has happened again with another file, though. I am worried, but I don't know how we should proceed if we are not able to reproduce.

See #407.

FelipeDefensor:
I am merging this before final approval to try and release v0.6.0 today. I don't think it will be possible though.

PR #391, opened 2026-03-10
Fix comments icon visibility

### LINKED ISSUE #345: Comments icon on hierarchy timeline is not being displayed
Hierarchies unit should have a icon on the top-right corner indicating they have comments. This icon is only being shown after we perform some operation like splitting.

The likely cause is that the apporpriate update function is not being called when we call `set_data` for the `comments` property.

FelipeDefensor:
Closes #345


azfoo (changes requested):
After changing the zoom level or resetting the start/end points, the icon disappears.

azfoo (approved):

PR #393, opened 2026-03-10
fix: selection box freezing when losing focus

FelipeDefensor:
This would happen when the user had an active selection box and would, for instance, receive an error message or tab out of the application.


azfoo (commented):
This solution should be extended to more dragging events (moving markers, slider, etc.). While it doesn't crash/freeze, it has the potential to cause erratic behaviour nonetheless.
Otherwise, this solves the described problem.

PR #398, opened 2026-03-16
feat: add title or file name to main window title

FelipeDefensor:
This is specially helpful when more than one TiLiA instance is running.


azfoo (commented):
I have a slight preference for `APP_NAME - title`, just so it's the first thing that comes up in case of a very long title. 
Also, having the title without the extension (default, media path), like when using the file path, would look a lot cleaner.

azfoo (approved):
I still have a slight preference for TiLiA - {some title}. Having the extension makes it clearer where the window gets its title from, and perhaps gives the user more reason to fill out the title field.

PR #399, opened 2026-03-16
fix: dynamically set audio outputs

FelipeDefensor:
Before this, changes to default audio outputs in the OS would be ignored. If the default output was removed, for instance, there was no way to regain sound playback, even if the output was reinserted.

This is the last one for v0.6, I promise.

### DIFF
--- a/tilia/media/player/qtplayer.py
+++ b/tilia/media/player/qtplayer.py
@@ -3,7 +3,13 @@ from __future__ import annotations
 import time
 
 from PyQt6.QtCore import QUrl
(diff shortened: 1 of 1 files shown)


FelipeDefensor:
I found no reasonable way to unit test this, although I didn't try that much. I have tested this manually for Windows. We should test this manually for Linux and Mac, too. @azfoo, could you do the Linux test?

Feel free to suggest postponing this to v0.6.1.

azfoo (approved):
Works on Linux, as far as I can tell.

PR #400, opened 2026-03-17
Fix empty media path being saved as "."

FelipeDefensor:
This lead to an error saying the media type/extension was unrecognised.

This should still go into v0.6 (sorry).


FelipeDefensor:
Created PR from wrong branch.

PR #404, opened 2026-03-17
feat: id validation

azfoo:
also removes scanning through all of the current ids if no id is given.

### DIFF
--- a/tilia/app.py
+++ b/tilia/app.py
@@ -281,10 +281,22 @@ class App:
         IDs are unique across timeline component and timelines.
         Other IDs might contain duplicates.
         """
+
+        def invalid_id(id):
+            tilia.errors.display(tilia.errors.INVALID_ID, id)
+            return str(next(self._id_counter))
+
         if id is None:
             return str(next(self._id_counter))
 
-        int_id = int(id)
+        if type(id) not in [int, str]:
+            return invalid_id(id)
+
+        try:
+            int_id = int(id)
+        except ValueError:
+            return invalid_id(id)
+
         timeline_ids = {int(c.id) for c in self.timelines}
         component_lists = [
             tl.components for tl in self.timelines if tl.components is not None
--- a/tilia/app.py
+++ b/tilia/app.py
@@ -275,28 +275,32 @@ class App:
             repeat_identifier=repeat_identifier,
         )
 
-    def get_id(self) -> str:
+    def get_id(self, id: str | None = None) -> str:
         """
         Returns an ID string.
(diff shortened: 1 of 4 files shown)


tilia/app.py
```diff
@@ -275,28 +275,32 @@ def on_record_state(self, action, no_repeat=False, repeat_identifier=""):
...
         Returns an ID string.
-        IDs are unique accross timeline component and timelines.
+        IDs are unique across timeline component and timelines.
         Other IDs might contain duplicates.
         """
-        timeline_ids = {c.id for c in self.timelines}
+        if id is None:
+            return str(next(self._id_counter))
+
+        int_id = int(id)
```
FelipeDefensor: This crashes if id is not parseable as an int. We should handle and test this possibility.

tilia/app.py
```diff
@@ -275,28 +275,32 @@ def on_record_state(self, action, no_repeat=False, repeat_identifier=""):
...
 
-        next_id = next(self._id_counter)
-        # Find the highest existing ID and increment until counter is higher
-        if existing_ids:
-            highest_id = max(int(id_str) for id_str in existing_ids)
-            while next_id <= highest_id:
-                next_id = next(self._id_counter)
-        return str(next_id)
+        if int_id in existing_ids:
+            return str(next(self._id_counter))
```
FelipeDefensor, line 310: `next(self._id_counter)` might already be in `existing_ids`, no?

tilia/app.py
```diff
@@ -275,28 +275,32 @@ def on_record_state(self, action, no_repeat=False, repeat_identifier=""):
...
+
+        if not existing_ids or int_id > max(existing_ids):
+            self._id_counter = itertools.count(int_id + 1)
```
FelipeDefensor, line 313: Much better than what we had.

FelipeDefensor (changes requested):
This is a clear improvement over my implementation. The only thing, apart from the comments, is that I am not sure we want to silently "adjust" conflicting IDs in the long term. A user might, for instance, write a script that creates a components with a hardcoded ID, and later use that ID to get the component for a future operation. That operation would misteriously fail if we changed the ID when creating the component without an error or warning. Since we are rather far from such an API in the moment, I am fine with going for what you propose, at least for now.

tilia/app.py
azfoo: No, see `test_create_component_with_higher_id_before_lower_id`. `self._id_counter` always increments from the largest input id, so there's no way it'll be smaller.

FelipeDefensor (approved):

PR #405, opened 2026-03-18
feat: displaying remote pdfs

azfoo:


FelipeDefensor (changes requested):
Is this meant for v0.6?

This is complex functionality, I think we need a bunch of tests. Some edge cases we need to handle/test:
- PDF is corrupted
- URL does not point to a file with the PDF extension
- Internet connection is down
- Request to download returns failure

I would bet there are some more. I imagine some of those are already handled, but without tests it is not so easy to tell.

Also, Python has the `tempfile` module to handle temporary files and directories. Have you considered using that?

I have not reviewed the code yet.

azfoo:
> Is this meant for v0.6?

Not sure, but it would probably be nice to have, especially when loading files directly off imslp or when downloading from the online database.

> Also, Python has the `tempfile` module to handle temporary files and directories. Have you considered using that?

The idea is that we save the files locally, but hidden, so that we keep the offline functionality after the first load, while keeping the original file links. There is a clear cache on exit option, but at the moment it doesn't delete the last opened pdfs, possibly because the files are in the process of being released by the closing app. I did consider tempfile, but that doesn't seem to persist between sessions.

FelipeDefensor:
>Is this meant for v0.6?
>Not sure, but it would probably be nice to have, especially loading files directly off imslp or when downloading from the online database.

Definitely. Let's schedule this for v0.6.x, then?

PR #415, opened 2026-03-25
fix: crash when using `timeline <kind> import` CLI command

FelipeDefensor:
The import logic was wrong, but tests didn't fail because the modules where already imported earlier in the session.

### DIFF
--- a/tilia/parsers/csv/__init__.py
+++ b/tilia/parsers/csv/__init__.py
@@ -1,0 +1,1 @@
+from . import hierarchy, marker, beat, harmony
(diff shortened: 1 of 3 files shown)


FelipeDefensor:
I don't know how we could test this automatically. @azfoo, could you reproduce and test the fix manually before approving?

azfoo:
Just realised I accidentally fixed this in 60e1629. Would there be any reason for choosing one over the other?

FelipeDefensor:
I am not entirely sure, but I think not. I am closing this.

PR #417, opened 2026-03-27
Darkmode

### LINKED ISSUE #413: Black notes and toolbar buttons over dark background when in dark mode
PySide apparently changes Qt color scheme according to the user's OS setting. While that works for most of the interface elements, the toolbar buttons and the notes in the SVG score viewer remain black, making them very difficult to read.

Image

azfoo:
closes #413 

(rebased on top of pyside fixes)


FelipeDefensor:
Hey, thanks for this. Why are the icons looking less sharp now?

grafik
Top image is before, bottom is after.

They feel a bit too low res for me, at the moment. Could we improve that?

azfoo:
> Hey, thanks for this. Why are the icons looking less sharp now?
> 
> 
> 
> grafik
> 
> Top image is before, bottom is after.
> 
> 
> 
> They feel a bit too low res for me, at the moment. Could we improve that?

I won't be able to do anything about the playback icons but I could make the other lines a bit thicker. Will update soon.
(Unless you're asking why the playback buttons look low res: they are greyed out because the buttons are disabled from having no media loaded.)

FelipeDefensor:
> I won't be able to do anything about the playback icons but I could make the other lines a bit thicker. Will update soon. (Unless you're asking why the playback buttons look low res: they are greyed out because the buttons are disabled from having no media loaded.)

The playback icons look great, thanks for switching to pre-made ones. The problem is not so much that the lines are thin (although thicker lines would help legibility), it is that the corners look jagged and the spaces between dots in the dotted lines uneven.

FelipeDefensor (approved):

PR #420, opened 2026-04-01
Cli improvements

azfoo:
mostly visual things

### DIFF
--- a/tilia/ui/cli/ui.py
+++ b/tilia/ui/cli/ui.py
@@ -25,9 +25,8 @@ from tilia.ui.cli import (
     script,
     timelines,
 )
-from tilia.ui.cli.io import ask_yes_or_no
+from tilia.ui.cli.io import ask_yes_or_no, tabulate
 from tilia.ui.cli.player import CLIVideoPlayer, CLIYoutubePlayer
-from tilia.constants import VERSION
 
 
 class CLI:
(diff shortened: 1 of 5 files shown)


tilia/ui/cli/ui.py
```diff
@@ -60,6 +61,12 @@ def setup_parsers(self):
         script.setup_parser(self.subparsers, self.parse_and_run)
         timelines.setup_parser(self.subparsers)
         self.subparsers.add_parser("about", help="About TiLiA").set_defaults(func=about)
+        self.parser.add_argument(
+            "--verbose",
+            "-v",
+            action=argparse.BooleanOptionalAction,
+            help="Toggle console logging",
+        )
```
FelipeDefensor: I don't think this should be called `verbose`, as the term implies that user-facing output will be generated, which is not not the case, currently. `logging` would be more appropriate. More importantly, are you sure that passing just something that looks like an option (that is, that starts with `--`) is idiomatic for CLIs? I am used to seeing commands like `timeline add hierarchy --verbose`, or `set-verbosity 1`, but not simply `--verbose`. Let's make this into a separate PR?

tests/ui/cli/test_cli_ui.py
```diff
@@ -33,3 +35,11 @@ def test_wrong_argument(self, cli, tilia_errors):
     def test_parse_command(self, cli, command, result):
 
         assert cli.parse_command(command) == result
+
+    def test_toggle_verbose(self, cli):
+        with patch("tilia.log.logger.on_settings_updated") as logger:
+            cli.parse_and_run("metadata set-media-length 60")
+            cli.parse_and_run("-v metadata set-media-length 60")
```
FelipeDefensor: Any reason for this not to be `metadata set-media-length 60 -v`?

FelipeDefensor (changes requested):
Apart from the verbosity thing, this looks good. Feel free to merge if you split that into a separate PR.

PR #422, opened 2026-04-17
feat: drag and drop to open file

azfoo:
Mostly because we currently have a slightly ridiculous process when trying to open an autosaved file


FelipeDefensor:
I would really like to have this, but it doesn't seem to work at all for me on Windows.

Edit: I think it works, but I found a bug.

FelipeDefensor (approved):
Review and test my fix and feel free to merge.

PR #441, opened 2026-04-20
fix: error message when scaling media & invalid media unloading previous

### LINKED ISSUE #425: Media load error is displayed when scaling components
See image

However, the media loads fine.

To reproduce
- Load media
- Create hierarchy timeline
- Load media with different duration
- Accept scale dialog

**n.b.** with a file that has media duration set (i.e. no actual media loaded), then has media loaded, the same error is displayed. Media in the backend seems to be loaded but the player controls are disabled.

### LINKED ISSUE #440: Loading invalid media unloads current media
Expected: error message is displayed and previous media remains loaded.
Actual: error message is displayed, no media is loaded.

FelipeDefensor:
@azfoo, I am not entirely sure how the `wait_for_signal` wrapper works, so take an extra careful look at this one.

The callback triggered by the media duration being available was being called before `wait_for_signal` when the wrapper resumed the event loop with `loop.exec()`. As the user takes some time to click the dialog, the wrapper ended up timing out and returning `False` as a success value, which triggered the error message.

This also somehow takes care of the previous valid media being unloaded when invalid files tried to be loaded.

I didn't find a way to manually test this. The whole media load process is fragile, so I suggest we redo manual tests before releasing.

Closes #440.
Closes #425.

### DIFF
--- a/tilia/media/player/qtplayer.py
+++ b/tilia/media/player/qtplayer.py
@@ -64,15 +64,11 @@ class QtPlayer(Player):
 
     def _init_player(self):
         self.player = QMediaPlayer()
-        self.player.durationChanged.connect(self.on_media_duration_available)
         self.player.setAudioOutput(self.audio_output)
 
     def on_media_load_done(self, path, start, end):
         super().on_media_load_done(path, start, end)
         post(Post.PLAYER_UPDATE_CONTROLS, PlayerStatus.PLAYER_ENABLED)
-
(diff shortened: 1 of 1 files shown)


FelipeDefensor:
Previous failure on CI was due to flaky test.

Note: `git commit --allow-empty -m "empty"` creates and empty commit that triggers CI.

azfoo (approved):

PR #442, opened 2026-04-20
Fix cli

### LINKED ISSUE #439: Event logs are printed to CLI console by default
To reproduce:
- Run TiLIA CLI from source with ENVIRONMENT=dev
- Open a file

Expected behaviour:
We get moderate text as feedback.

Actual behaviour:
Every event gets printed to the console.

Having the events printed is useful for developing, but I feel the default should be that those logs do not get printed, as running the CLI requires that regular users run TiLiA from source.

azfoo:
closes #439

### DIFF
--- a/tests/ui/cli/test_cli_ui.py
+++ b/tests/ui/cli/test_cli_ui.py
@@ -1,3 +1,5 @@
+from unittest import mock
+
 import pytest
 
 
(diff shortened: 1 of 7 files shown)


FelipeDefensor (changes requested):
Quitting with `exit` or `quit` is not working anymore. Also, triggering a keyboard interrupt does not exit immediately, only after the next command is input and confirmed with enter. Did you test this manually?

azfoo:
> Quitting with `exit` or `quit` is not working anymore. Also, triggering a keyboard interrupt does not exit immediately, only after the next command is input and confirmed with enter. Did you test this manually?

Sorry, got over-excited. `quit` is fixed now. 
Don't think it is possible to have KeyboardInterrupt be caught if it is the first thing that input sees without enter. (i.e. `[interrupt] [enter]`, `[some other string] [interrupt] (without enter)`, and `[interrupt] [something else] [enter]` will quit but not `[interrupt]` alone.) Probably something to do with how input blocks the signal. There is going the long way round and building a custom parser, but nonetheless, I don't think a KeyboardInterrupt should register as a crash.

FelipeDefensor (approved):
Nice bonus is that KeyboardInterrupt doesn't create a crash dialog in the GUI anymore.

Issue #429, opened 2026-04-20
none of the fields in harmony inspector changes anything


FelipeDefensor:
Expected: changing the value changes the harmony

PR #446, opened 2026-04-22
fix: inspector combobox using wrong values

azfoo:
closing #429

### DIFF
--- a/tests/ui/dialogs/test_inspect.py
+++ b/tests/ui/dialogs/test_inspect.py
@@ -1,6 +1,10 @@
+import pytest
+
 from tests.conftest import parametrize_ui_element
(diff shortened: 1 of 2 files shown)


FelipeDefensor (approved):
This makes sense, but I wasn't able to reproduce #429. What do you meant with "None of the fields in the harmony inspector change anything"? They seemed to already work fine for me.

azfoo:
> This makes sense, but I wasn't able to reproduce #429. What do you meant with "None of the fields in the harmony inspector change anything"? They seemed to already work fine for me.

This might be a setup-dependent problem; the inspector was returning the index of the combo box rather than its data, so set data failed validation and resulted in the value not changing.

PR #447, opened 2026-04-22
Misc fixes

### LINKED ISSUE #428: 'QGraphicsItem.paint' not implemented when playing with score
To reproduce
- Load audio
- Create Score timeline
- Create beat timeline
- Fill beat timeline with enough beats
- Load score
- Seek to an arbitrary time (before playing)
- Play audio

Error message
```
Traceback (most recent call last):
  File "C:\Users\Felipe\dev\tilia-desktop\tilia\media\player\base.py", line 249, in _play_loop
    self.current_time = self._engine_get_current_time() - self.playback_start
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Felipe\dev\tilia-desktop\tilia\media\player\qtplayer.py", line 96, in _engine_get_current_time
    return self.player.position() / 1000
           ^^^^^^^^^^^^^^^^^^^^
NotImplementedError: pure virtual method 'QGraphicsItem.paint' not implemented.
```

Notes
This does not seem to happen if we start playing from the start of the song.

azfoo:
29177d3 seems to resolve #428

### DIFF
--- a/tilia/ui/timelines/hierarchy/timeline.py
+++ b/tilia/ui/timelines/hierarchy/timeline.py
@@ -297,16 +297,14 @@ class HierarchyTimelineUI(TimelineUI):
             get(Get.TIMELINE_COLLECTION).set_timeline_data(
                 self.id, "height", max_height + self.MIN_MARGIN
             )
-        return True
 
     @with_elements
     def on_increase_level(self, elements: list[HierarchyUI]) -> bool:
-        return (
-            self.timeline.alter_levels(
-                self.elements_to_components(list(reversed(elements))), 1
-            )
-            and self._adjust_timeline_height()
-        )
+        if success := self.timeline.alter_levels(
+            self.elements_to_components(list(reversed(elements))), 1
+        ):
+            self._adjust_timeline_height()
+        return success
 
     @with_elements
     def on_decrease_level(self, elements: list[HierarchyUI]):
@@ -314,10 +312,9 @@ class HierarchyTimelineUI(TimelineUI):
 
     @with_elements
     def on_group(self, elements: list[HierarchyUI]):
-        return (
-            self.timeline.group(self.elements_to_components(elements))
-            and self._adjust_timeline_height()
-        )
+        if success := self.timeline.group(self.elements_to_components(elements)):
+            self._adjust_timeline_height()
+        return success
 
     def on_split(self, time: float | None = None):
         if time is None:
--- a/tilia/ui/timelines/hierarchy/timeline.py
+++ b/tilia/ui/timelines/hierarchy/timeline.py
@@ -178,17 +178,13 @@ class HierarchyTimelineUI(TimelineUI):
         prev_parent_length = prev_parent_end - prev_parent_start
         scale_factor = new_parent_length / prev_parent_length
 
-        relative_child_start = (
-            child_pastedata_["support_by_component_value"]["start"] - prev_parent_start
-        )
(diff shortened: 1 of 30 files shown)


tilia/ui/timelines/hierarchy/timeline.py
```diff
@@ -294,28 +295,33 @@ def on_paste_element_complete(self, clipboard_contents: dict) -> bool:
...
     @with_elements
     def on_increase_level(self, elements: list[HierarchyUI]) -> bool:
-        min_margin = 10
-        success = self.timeline.alter_levels(
-            self.elements_to_components(list(reversed(elements))), 1
+        return (
+            self.timeline.alter_levels(
+                self.elements_to_components(list(reversed(elements))), 1
+            )
+            and self._check_timeline_height()
```
FelipeDefensor: I don't see why _check_timeline_height would return a bool if we didn't need to make this line work. This is conciser, but the following would be more readable, and do without the _check_timeline_height return value:

```
result = self.timeline.alter_levels(...)
self._check_timeline_height()
return result
```

tilia/ui/timelines/hierarchy/timeline.py
```diff
@@ -294,28 +295,33 @@ def on_paste_element_complete(self, clipboard_contents: dict) -> bool:
...
     def on_decrease_level(self, elements: list[HierarchyUI]):
         return self.timeline.alter_levels(self.elements_to_components(elements), -1)
 
     @with_elements
     def on_group(self, elements: list[HierarchyUI]):
-        return self.timeline.group(self.elements_to_components(elements))
+        return (
+            self.timeline.group(self.elements_to_components(elements))
+            and self._check_timeline_height()
+        )
```
FelipeDefensor: Same as above.

tilia/ui/timelines/hierarchy/timeline.py
```diff
@@ -294,28 +295,33 @@ def on_paste_element_complete(self, clipboard_contents: dict) -> bool:
 
         return True
 
+    def _check_timeline_height(self):
```
FelipeDefensor: This does more than check. A better name would be `_adjust_timeline_height`.

FelipeDefensor (commented):
This is mostly approved. There is just one relevant bug for which I am working on a fix. Will push it tomorrow.

It would be good to address the comments, but it is not priority.

tilia/ui/timelines/hierarchy/timeline.py
azfoo: > I don't see why _check_timeline_height would return a bool if we didn't need to make this line work. This is conciser, but the following would be more readable, and do without the _check_timeline_height return value:
> 
> ```
> result = self.timeline.alter_levels(...)
> self._check_timeline_height()
> return result
> ```

Because doing `return a() and b()` means that if `a()` returns False, `b()` doesn't run and False is returned. Still, happy to rewrite.

FelipeDefensor: Oh, that's true, but the performance gain from not evaluating `b()` is negligible in this case, right? In any case, we could still do:
```
result = self.timeline.alter_levels(...)
if result:
    self._check_timeline_height()
return result
```

FelipeDefensor (approved):
Good to go from my part. Addressing the comments is optional. Review my changes before merging.

FelipeDefensor (commented):
The capital letter already indicated the default on your implementation. I forgot about that convention, sorry. At least we have the option of setting a default now.

PR #449, opened 2026-04-24
Youtube queue

### LINKED ISSUE #392: Seeking before YouTube video is loaded displays error
Opening a file and imediately trying to seek to another time may cause this error be displayed:

Image

It seems it happens when video hasn't finished loading before the seek attempt. The video still loads fine afterwards.

### LINKED ISSUE #437: Error when opening file with YouTube stream loaded
See
We get this log message: QtCriticalMsg: Uncaught TypeError: player.loadPlaylist is not a function
To reproduce
- open file with YouTube video loaded
- open file with audio loaded
- open file with YouTube video loaded

azfoo:
closes #437

### DIFF
--- a/tilia/media/player/youtube.html
+++ b/tilia/media/player/youtube.html
@@ -87,10 +87,6 @@
         return player.getDuration()
       }
 
-      function getCurrentTime() {
-        return player.getCurrentTime()
-      }
-
       function play() {
         return player.playVideo()
       }
--- a/tilia/media/player/youtube.html
+++ b/tilia/media/player/youtube.html
@@ -30,13 +30,8 @@
       const firstScriptTag = document.getElementsByTagName("script")[0];
       firstScriptTag.parentNode.insertBefore(tag, firstScriptTag);
 
-      function _fnParams(fn, ...params) {
-        this.fn = fn;
-        this.params = params;
-      }
-
       function _fnCall(fp) {
-        player[fp.fn](fp.params);
+        player[fp.fn](...fp.params);
       }
 
       function _deQueue() {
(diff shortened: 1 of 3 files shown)


FelipeDefensor (commented):
AI noticed this breaks other player behavior:

**File:** `tilia/media/player/youtube.html`

The chain of `tryPlayerFunctionElseQueue` → `new _fnParams` → `_fnCall` wraps arguments in an extra array at every step.

```js
// _fnParams uses rest params, so it wraps its args:
function _fnParams(fn, ...params) { this.params = params; }
// tryPlayerFunctionElseQueue also uses rest params, collecting into an array:
function tryPlayerFunctionElseQueue(fn, ...params) {
    const fp = new _fnParams(fn, params);  // <-- passes array as a single arg
```

When `tryPlayerFunctionElseQueue("seekTo", time, true)` is called:
- `params = [time, true]` (rest collects)
- `new _fnParams("seekTo", [time, true])` — the array is passed as one argument
- Inside `_fnParams`, rest again wraps it: `fp.params = [[time, true]]`
- `_fnCall(fp)` → `player["seekTo"]([[time, true]])` ← Wrong. Should be `player.seekTo(time, true)`.

For `loadPlaylist` (the primary fix in this PR), the nesting is even deeper. `loadVideo` calls `tryPlayerFunctionElseQueue("loadPlaylist", [id])`, resulting in `player.loadPlaylist([[[id]]])` — triple-nested — instead of `player.loadPlaylist([id])`.

All calls that pass arguments are broken: `seekTo`, `setVolume`, `setPlaybackRate`, `setLoop`, and `loadPlaylist`. No-argument calls (`playVideo`, `pauseVideo`, `stopVideo`, `mute`, `unMute`) accidentally work because extra JS args are ignored.

**Fix** — replace the constructor pattern with a plain object and spread on call:

```js
// Remove _fnParams constructor entirely. Replace:
const fp = new _fnParams(fn, params);
// with:
const fp = { fn, params };  // params is already the correct array
// And in _fnCall, spread the args:
function _fnCall(fp) {
    player[fp.fn](...fp.params);
}
```

I will push the fix right away. I want to be extra thorough on manual testing after this, since we have no automated tests for any of our JS.

FelipeDefensor (approved):

PR #450, opened 2026-04-24
Youtube extras

azfoo:


FelipeDefensor:
Hey, is this meant for v0.6.0?

azfoo:
In theory, yes.

FelipeDefensor:
Would it be too bad if we left it for v0.6.1? I think we have enough on our plates if we are releasing tomorrow, especially because I am calling it a day.

azfoo:
Not particularly bad. I'd like 79cca23 to go out for 0.6, mostly because its annoying showing the world your yt watch history, (and maybe 411495e because that's fairly simple). No worries if not.

(in any case, I'm going to trigger the build process and we can start the testing tomorrow)

FelipeDefensor (changes requested):
Some issues with this:
- Fullscreen doesn't work at all for me (maybe we talked about this and I forgot, sorry).
- Autoplay only works if I clicked play for a previous video in the same TiLiA session.
- icecream is not listed as a dependency, if I remember correcly, so `ic` has a potential to crash. Replace with another logging mechnism.
- If I try to load a well-formed Youtube URL with an invalid video ID, the video is not loaded, but I get no error message at all.

Also, I got this review comment from Claude, it seems to make sense, but I haven't been able to test it.

- display_error semantics changed for all callers. It's reached from PlayerTracker.on_error (line 62), which forwards every YT JS onError (invalid video ID, embed disabled by uploader, region-restricted,
  etc.). A reload won't recover from any of those — but the user now gets an unsolicited "Would you like to try and refresh the page?" prompt on every JS error. Consider gating the refresh prompt on
  recoverable errors only, or at minimum keep the error-surface call (so the failure still flows through tilia.errors) and offer reload as a secondary step.

FelipeDefensor:
Is this ready for review again? Is this for 0.7?

azfoo:
yes I think so

FelipeDefensor (approved):
Significant improvement to UX. Thanks for the features and all the revisions!

I have added a small fix as a last commit.

During testing I found a lot of related issues to improve, and opened #590 and #591 accordingly. Maybe we can schedule them for v0.7.1.

PR #458, opened 2026-04-27
Tests for our JavaScript

### LINKED ISSUE #459: There are no automated tests for any of our JavaScript

FelipeDefensor:
This is the draft of a test suite for our JavaScript created by Claude Opus 4.6 on maximum effort. This seems very convoluted, and I wonder if it can't be simplified. I am pushing this only for future reference.

Depends on #449.


FelipeDefensor:
We have a lot of open PRs, and I bet we can do better than this.

PR #461, opened 2026-04-27
ci: rerun tests if the first try doesn't work

azfoo:


.github/workflows/run-tests.yml
```diff
@@ -80,7 +80,9 @@ jobs:
 
     - name: Test with pytest
       run: |
-        coverage run -m pytest -v -s
+        function run(){ coverage run -m pytest -vs; }
+        run || run
```
FelipeDefensor: Some useful AI feedback:
  - run || run will surface failures if both runs fail — that part is correct.
  - However, retrying the entire suite masks flaky tests. A test that fails ~40% of the time will appear green in CI indefinitely.
  - For Qt/GUI test suites, some flakiness is unavoidable (race conditions, off-screen rendering timing), so a retry can be
  pragmatically justified — but it should be scoped as narrowly as possible, e.g. using pytest-rerunfailures to retry only individual
   failing tests rather than the whole suite.
  - In the worst case (consistent failure) this doubles CI time before surfacing the error.
  - Coverage data: if the first run fails mid-suite, the second run overwrites .coverage, so the coverage report is only from the
  second run. This could silently under-report coverage.

  Suggestion: Use pytest-rerunfailures (--reruns 1 --reruns-delay 1) instead, which retries only the individual failing test(s), keeps coverage accurate, and makes it visible which tests are flaky.

pyproject.toml
```diff
@@ -94,6 +94,7 @@ env = [
     "QT_QPA_PLATFORM=offscreen",
 ]
 timeout = 10
+filterwarnings = ["ignore::pytest.PytestConfigWarning"]
```
FelipeDefensor: Why are we doing this? We need a comment here.

FelipeDefensor (changes requested):

PR #480, opened 2026-04-28
fix: use current measure number as initial value in set measure number dialog

### LINKED ISSUE #471: Initial value for set measure number dialog should probably not be 0
The actual measure number is a better candidate.

FelipeDefensor:
## Summary
- Prefills the \"Set measure number\" dialog with the measure number of the first selected beat, instead of always defaulting to 0.

## Test plan
- [ ] Right-click a beat, select \"Set measure number\" — dialog should open pre-filled with the current measure number for that beat.

Closes #471


FelipeDefensor:
Add tests for this.

PR #481, opened 2026-04-28
fix: close SVG viewer and clear svg_data when score timeline is cleared

### LINKED ISSUE #472: Clearing a score timeline does not close the SVG viewer
This looks harmless enough, but I think it might lead to bugs if the app continues to be used. What happens, for instance, when we try to save the file?

EDIT: Said and done, I continued messing with the file and eventually got a crash.

FelipeDefensor:
## Summary
- Clears `svg_data` in the backend `ScoreTLComponentManager.clear()` so the `svg_view` property cannot recreate a viewer with stale SVG data after clearing.
- Fixes `reset_svg()` to access the existing viewer directly via `get()` instead of going through the `svg_view` property (which auto-creates a viewer if none exists).

## Test plan
- [x] Create a score timeline and import a MusicXML file — SVG viewer opens.
- [x] Clear the score timeline (Edit > Clear) — SVG viewer should close and not reappear.
- [x] Continue using the app — no crash when saving or performing other operations.

Closes #472

### DIFF
--- a/tilia/timelines/score/timeline.py
+++ b/tilia/timelines/score/timeline.py
@@ -59,6 +59,7 @@ class ScoreTLComponentManager(TimelineComponentManager):
         post(Post.SCORE_TIMELINE_COMPONENTS_DESERIALIZED, self.timeline.id)
 
     def clear(self):
+        self.timeline.save_svg_data("")
         super().clear()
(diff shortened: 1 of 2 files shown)


FelipeDefensor:
Add tests for this.

PR #483, opened 2026-04-28
feat: add harmony and audiowave timeline kinds to CLI add command

### LINKED ISSUE #476: Not all timeline kinds can be created with CLI
Only hierarchy, marker, beat and score timelines can be created at the moment.

```
>>> tl add -h
usage: TiLiA timelines add [-h] [--name NAME] [--height HEIGHT] [--beat-pattern BEAT_PATTERN [BEAT_PATTERN ...]] {hierarchy,hrc,marker,mrk,beat,bea,score,sco}

positional arguments:
  {hierarchy,hrc,marker,mrk,beat,bea,score,sco}
                        Kind of timeline to add
```

FelipeDefensor:
## Summary
- Adds `harmony` / `har` and `audiowave` / `aud` as valid kind arguments to `timelines add`.
- Harmony timelines omit the `--height` kwarg (their height is computed from level count and level height internally).
- Audiowave timelines similarly omit `--height`.
- Tests added for both new kinds and their abbreviations.

## Test plan
- [ ] `tilia timelines add harmony --name "Chords"` creates a harmony timeline.
- [ ] `tilia timelines add har` works as shorthand.
- [ ] `tilia timelines add audiowave --name "Wave"` creates an audiowave timeline.
- [ ] `tilia timelines add aud` works as shorthand.

Closes #476

### DIFF
--- a/tests/ui/cli/timelines/test_cli_timelines_add.py
+++ b/tests/ui/cli/timelines/test_cli_timelines_add.py
@@ -59,3 +59,27 @@ class TestTimelineAdd:
 
         tl = tls.get_timelines()[0]
         assert tl.beat_pattern == [4]
+
+    def test_add_harmony_timeline(self, cli, tls):
+        cli.parse_and_run("timeline add har --name test")
+
+        tl = tls.get_timelines()[0]
+        assert tl.KIND == TimelineKind.HARMONY_TIMELINE
+        assert tl.name == "test"
+
+    def test_add_harmony_timeline_full_name(self, cli, tls):
+        cli.parse_and_run("timeline add harmony")
+
+        assert tls.get_timelines()[0].KIND == TimelineKind.HARMONY_TIMELINE
+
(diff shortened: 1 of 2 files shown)


FelipeDefensor:
Superseded by #575.

This branch was cut ~209 commits before current `dev` and no longer merges — the `TimelineKind` enum it's written against has since been removed in favour of `Timeline` subclasses, and `add.py` was restructured around that.

Review also turned up that the `audiowave` kind could not have worked from the CLI at all: `create_timeline` calls `AudioWaveTimeline.refresh()`, which requests `Get.PLAYBACK_AREA_WIDTH` — served only by `QtUI`. With media loaded it raised `NoReplyToRequest` and left an empty timeline behind; without media it produced a hidden, empty one. The tests here passed because they never loaded media and asserted only kind and name, both of which hold for a hidden empty timeline.

#575 reimplements this on `dev` with that fixed, adds `pdf` so #476 actually closes, and rejects `--height` for harmony instead of silently dropping it.

PR #485, opened 2026-04-29
fix: zip mac .app using relative path to avoid nested directory structure in archive

### LINKED ISSUE #477: Mac executable is inside multiple nested directories
Image

FelipeDefensor:
## Summary

- Fixes the Mac distributable containing the full runner path (`Users/runner/work/desktop/desktop/build/macos-silicon/exe/tilia.app`) inside the zip instead of just `tilia.app` at the root.
- Root cause: `zip` was called with an absolute path for the `.app` bundle, causing the entire directory hierarchy to be embedded in the archive.
- Fix: run the `zip` command from the parent directory of the `.app` (using a subshell `cd`) so only the relative name `tilia.app` appears at the root of the zip.

## Test plan

- [x] Trigger a Mac build via workflow dispatch and verify the downloaded zip extracts `tilia.app` directly at the root with no nested directories.

Closes #477

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -164,7 +164,7 @@ jobs:
           ls build -ltraR |egrep -v '\.$|\.\.|\.:|\.\/|total|^d' | sed '/^$/d' || true
 
       - name: Upload
-        uses: "softprops/action-gh-release@v2"
+        uses: "softprops/action-gh-release@v3"
         id: release
         with:
           tag_name: ${{github.ref_name}}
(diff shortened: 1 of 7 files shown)


FelipeDefensor:
This results in a better, but still wrong structure: `TiLiA-v0.6.0-<tag>-macos-silicon.zip -> macos-silicon/exe -> TiLiA-v0.6.0-<tag>-macos-silicon.zip -> tilia.app`.

Since verifying this takes so long, I would say we release v0.6.0 with this and fix it for v0.6.1. @azfoo, what do you think?

Actually, we can manually edit the release and add the right binary `TiLiA-v0.6.0-<tag>-macos-silicon.zip`, so one more reason to go ahead with the release.

azfoo:
Hang on, will update in 20 mins.
(Still waiting, sorry)

azfoo:
22f90a2 fixes 9bd3369, 03ec2b9 cherry-picked to dev.

Issue #384, opened 2026-03-04
Order of elements in JSON changes in a new session


johentsch:
In the first session, I had done a couple of changes, reviewing a form analysis. Before committing each change to git, I exported the JSON file to produce a readable diff. Each time, the JSON would display exactly the changes applied to the analysis.

In the second session, i.e. after sting TiLiA once again, I only corrected a small spelling mistake but this time, when I stored the JSON, the order of a few elements changed. The result can be viewed in this comment: [a1df142](https://github.com/MusicalForm/annotation_pilot/pull/43/commits/a1df1421094a82051f1fdd66675abd80e65ceff3) (private repo).

FelipeDefensor:
I pushed an experimental fix to the `lcma` branch. It attempts to enforce a consistent order to the dictionary that is used both for the TiLiA file and for the JSON export. Any first save after this should contain **a lot** of changes, but be stable afterwads. I have commited a file after the first save to [this PR](https://github.com/MusicalForm/annotation_pilot/pull/83) (private repo) in the annotation pilot. Could you try opening it with TiLiA with the changes in the lcma branch and checking if doing an atomic changes and saving/export to JSON yields a sane diff?

johentsch:
Thanks for the quick fix! I can confirm it solved the issue in this case.

PR #498, opened 2026-05-03
feat: persist ids and components order when saving

FelipeDefensor:
Saved files now preserve component IDs and ordering, so reloading produces a deterministic structure. This is fundamental for tidy version control of TiLiA files.

### DIFF
--- a/tests/timelines/test_serialize.py
+++ b/tests/timelines/test_serialize.py
@@ -1,0 +1,345 @@
+import json
+from pathlib import Path
+
+from tests.mock import Serve, patch_yes_or_no_dialog
+from tests.ui.timelines.marker.interact import click_marker_ui
+from tests.utils import save_and_reopen, save_tilia_to_tmp_path, undoable
+from tilia.requests import Get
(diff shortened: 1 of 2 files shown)


FelipeDefensor:
Closes #384

tests/timelines/test_serialize.py
```diff
@@ -0,0 +1,345 @@
...
+    def test_preserves_ids_in_dict(self, marker_tl):
+        m1, _ = marker_tl.create_component(ComponentKind.MARKER, time=0)
+        m2, _ = marker_tl.create_component(ComponentKind.MARKER, time=1)
+
+        result = serialize_components(marker_tl.components)
+
+        assert set(result.keys()) == {m1.id, m2.id}
+
+
+class TestSerializeDeserializeRoundTrip:
```
azfoo, line 38: maybe TestSequentialSerializeDeserialize?

azfoo (approved):
Approving, but I still think there must be a better name to describe what we're testing

PR #500, opened 2026-05-03
chore: misc non-range cleanup

FelipeDefensor:
12 commits with small fixes/improvements. Including:
- `refactor: return success from App.on_open`
- `fix: inheritance/menu-name, typing`
- `chore: take path parameter to file.save_as command`
- `chore: pass y argument to timelines context menus`
- `feat: skip QGraphicsItems with ignore_right_click=True on right click`
- `fix: rename "Set name" timeline action to "Set timeline name"`
- `chore: tidy debug print and GetComponentDataError message`
- `fix: avoid invalid-return-value error when set_timeline_data is a no-op`
- `fix: block widget signals during programmatic value updates`
- `feat: rename and relocate "Move timeline up/down"`
- `test: cover slider click + drag-release seeking media`

### DIFF
--- a/tilia/ui/timelines/base/context_menus.py
+++ b/tilia/ui/timelines/base/context_menus.py
@@ -1,7 +1,14 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
 from tilia.requests import Get, get
 from tilia.ui import commands
 from tilia.ui.commands import CommandQAction, get_qaction
 from tilia.ui.menus import MenuItemKind, TiliaMenu
+
+if TYPE_CHECKING:
+    from tilia.ui.timelines.base.timeline import TimelineUI
 
 
 class TimelineUIContextMenu(TiliaMenu):
@@ -11,7 +18,7 @@ class TimelineUIContextMenu(TiliaMenu):
         (MenuItemKind.COMMAND, "timeline.set_height"),
     ]
 
-    def __init__(self, timeline_ui):
+    def __init__(self, timeline_ui: TimelineUI, x: int, y: int):
         super().__init__()
         self.timeline_ui = timeline_ui
         self._add_timeline_actions()
--- a/tilia/ui/timelines/base/context_menus.py
+++ b/tilia/ui/timelines/base/context_menus.py
@@ -24,8 +24,8 @@ class TimelineUIContextMenu(TiliaMenu):
         self._add_timeline_actions()
 
     def _add_timeline_actions(self):
-        self.addSeparator()
-
+        # Move up/down belong with the top "this timeline" actions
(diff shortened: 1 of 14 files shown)


TESTING.md
```diff
@@ -46,5 +46,14 @@ You will find many examples of the former in the test suite, though. Refactors a
...
 
+## Debugging bugs that only reproduce in the running app
+
+Sometimes a bug is reported in TiLiA but the test suite stays green — typically because something in the live path (a Qt signal, a focus change, the inspector dock pushing a stale snapshot back to a component) isn't exercised by the test harness. When this happens — or when you're working with an LLM agent that can't drive the GUI itself — the fastest way forward is to instrument the suspect code path and reproduce live:
+
+- Add `print("[FEATURE-DEBUG] ...", flush=True)` lines at points where the data flow is unclear: inside the command handler, around `set_component_data` calls, inside `Post.*` listeners that might overwrite state. Use a unique tag per investigation so the prints are trivial to grep and strip.
+- Run TiLiA from a terminal you can read, perform the exact action, and inspect the tagged lines.
+- Once the cause is identified, remove the prints in the same commit as the fix — don't ship debug output.
+- Prefer this over guessing when tests cover the symptom but the bug still reproduces. The instrumentation almost always pinpoints a side effect of a Qt signal or listener you didn't account for.
+
```
FelipeDefensor: These instructions should live only in CLAUDE.md. If they are not there, make a commit and add it to the PR that introduces it.

tilia/ui/timelines/base/context_menus.py
```diff
@@ -11,14 +18,14 @@ class TimelineUIContextMenu(TiliaMenu):
         (MenuItemKind.COMMAND, "timeline.set_height"),
     ]
 
-    def __init__(self, timeline_ui):
+    def __init__(self, timeline_ui: TimelineUI, x: int, y: int):
```
azfoo, line 21: am I right to assume that this actually gets used somewhere later on?

tilia/ui/timelines/view.py
```diff
@@ -65,12 +65,18 @@ def handle_left_click():
             )
 
         def handle_right_click():
+            items = [
+                item
+                for item in self.items(event.pos())
+                if not getattr(item, "ignore_right_click", False)
```
azfoo, line 71: what has the attribute "ignore_right_click"?

azfoo (changes requested):

tilia/ui/timelines/base/context_menus.py
FelipeDefensor: Yes, range timelines will use it. Maybe it would have made more sense to leave this change in that commit.

PR #499, opened 2026-05-03
refactor: get rid of TimelineKind

FelipeDefensor:
Removes the `TimelineKind` enum in favor of direct Timeline subclass references throughout command/menu registries and timeline lookups. This leverages the fact that, in Python, classes are first-class citizens and reduces duplication. It also makes it easier to introduce new timeline types, as less code that is outside the timeline module needs to be touched.

This was done in preparation for range timelines, and the range timeline PR (#501) is stacked on top of this one.

PR #497, opened 2026-05-03
docs: track CLAUDE.md

FelipeDefensor:
I have used Claude 4.7 heavily for developing the range timeline and it sped up my work immensely.

PR #387, opened 2026-03-07
Media commands

FelipeDefensor:
Also add tests for some of the related behaviour.

PR #501, opened 2026-05-03
feat: range timeline

FelipeDefensor:
## Summary

Adds the **range timeline** — a new timeline kind for annotating arbitrary time ranges, organised into rows. Ranges can be joined into chains, merged, split, and extended with hierarchy-style pre-start / post-end whiskers.

## What's in scope

**UI**
- Element / timeline / row context menus
- Ribbon-style toolbar grouped into Ranges / Rows / Display sections, with split-mode dropdown
- Inspector with label, start, end, length, pre-start/post-end, comments
- Comments indicator on the body (visible when comments are non-empty and the body is wide enough)
- Vertical arrow navigation; horizontal + vertical body drag (snaps to row)
- Copy/paste, full undo/redo

**CSV import (by-time + by-measure)**
- Required: `start`, `end`, `row`. Optional: `label`, `color`, `comments`, `joined_with_next`, `start_fraction`, `end_fraction`
- Rows auto-created by name; joins validated after parsing
- Replaces existing rows; restores >=1-row invariant on empty import

**CLI**
- `timelines add range [--row-height N]`
- `timelines range row {add, set-height, list, rename, set-color, reset-color, reorder, remove}`
- `timelines import range {by-time, by-measure}`

**File format**
- `RangeTimeline.SERIALIZABLE`: `rows`, `default_row_height`
- `Range` adds `row_id`, `label`, `color`, `comments`, `joined_right`, `pre_start`, `post_end`, `start_fraction`, `end_fraction`
- Per-row `Row.height` (optional override)
- Label alignment is session-only (settings, not per-timeline)

## Stack

This branch sits on top of the cleanup stack (#497, #498, #499, #500) plus the media-commands refactor (#387). When those land on `dev`, the merge commits drop on rebase via patch-id matching and the squash sits cleanly on top. No further history rewriting required here.

## Tests

- 407 range-specific tests across `tests/timelines/range/`, `tests/ui/timelines/range/`, `tests/parsers/csv/test_range_from_csv.py`, `tests/ui/cli/timelines/test_cli_timelines_range.py` and `test_cli_timelines_import.py`
- Full suite: 1612 passed, 0 failed


examples/range/by_measure_basic.csv
FelipeDefensor, line 1: - Examples folder should not be in this repo, but on the testing repo
- Also, I am not able to open files from the testing repo, open a issue there mentioning

tests/parsers/csv/test_range_from_csv.py
```diff
@@ -0,0 +1,353 @@
...
+
+
+def _import_by_time(timeline, data):
+    with patch("builtins.open", mock_open(read_data=data)):
+        return import_by_time(timeline, Path())
+
+
+def _import_by_measure(timeline, beat_tl, data):
+    with patch("builtins.open", mock_open(read_data=data)):
+        return import_by_measure(timeline, beat_tl, Path())
```
FelipeDefensor: Aren't these redundant with the helper function below? If so, we should prefer using that helper, as it replicates production more closely.

tests/parsers/csv/test_range_from_csv.py
```diff
@@ -0,0 +1,353 @@
...
+        assert "Verses" in names
+        assert "Choruses" in names
+
+    def test_reuses_existing_row(self, range_tl):
+        existing = range_tl.rows[0]
+        existing.name = "Existing"
+        data = "start,end,row\n0,1,Existing\n1,2,Existing"
+        _import_by_time(range_tl, data)
+        # Only one row should exist (no duplicate "Existing").
+        assert sum(1 for r in range_tl.rows if r.name == "Existing") == 1
```
FelipeDefensor: Could be len(r for r in range_tl.rows if r.name == "Existing")

tests/parsers/csv/test_range_from_csv.py
```diff
@@ -0,0 +1,353 @@
...
+
+    def test_with_fractions(self, range_tl, populated_beat_tl):
+        data = "start,end,start_fraction,end_fraction,row\n1,2,0.5,0.5,A"
+        _import_by_measure(range_tl, populated_beat_tl, data)
+        assert len(range_tl) == 1
+        # Beats at 1,2 → measure 1 starts at 1, measure 2 starts at 2;
+        # fraction 0.5 within measure 1 (length 1) gives 1.5; within
+        # measure 2 (length 1, end-side) gives 2.5.
+        ranges = list(range_tl)
+        assert ranges[0].start == 1.5
```
FelipeDefensor: We can index range_tl directly, we do not need the intermediary conversion to list.

tests/parsers/csv/test_range_from_csv.py
```diff
@@ -0,0 +1,353 @@
...
+            "0,1,A,true\n"
+            "0,1,B,false\n"
+            "1,2,A,false"
+        )
+        _import_by_time(range_tl, data)
+        row_a = next(r for r in range_tl.rows if r.name == "A")
+        row_b = next(r for r in range_tl.rows if r.name == "B")
+        a_ranges = sorted(
+            (r for r in range_tl if r.row_id == row_a.id), key=lambda r: r.start
+        )
```
FelipeDefensor: This seems like a common enough pattern that we would want to have range_tl.get_ranges_by_row method in RangeTimeline. Check the rest of the codebase to see if it is worth it.

tests/parsers/csv/test_range_from_csv.py
```diff
@@ -0,0 +1,353 @@
...
+
+class TestJoinValidation:
+    def test_flag_on_last_range_reports_error(self, range_tl):
+        data = "start,end,row,joined_with_next\n0,1,A,false\n1,2,A,true"
+        success, errors = _import_by_time(range_tl, data)
+        assert success
+        assert any("last range" in e and "'A'" in e for e in errors)
+        # Ranges still created; only the join is dropped.
+        assert len(range_tl) == 2
+        assert all(r.joined_right is None for r in range_tl)
```
FelipeDefensor: range_tl[-1].joined_right is simpler access

tests/timelines/range/test_range_timeline.py
```diff
@@ -0,0 +1,220 @@
...
+        )
+        r2, _ = range_tl.create_component(
+            ComponentKind.RANGE, id=2, start=10, end=20, row_id=row_id
+        )
+
+        serialized_components = range_tl.component_manager.serialize_components()
+
+        assert serialized_components[r1.id]["start"] == r1.start
+        assert serialized_components[r1.id]["end"] == r1.end
+        assert serialized_components[r1.id]["row_id"] == r1.row_id
```
FelipeDefensor, line 34: Also assert the second components properties.

tests/timelines/range/test_range_timeline.py
```diff
@@ -0,0 +1,220 @@
...
+        )
+        range_tl.create_component(
+            ComponentKind.RANGE, id=2, start=10, end=20, row_id=row_id
+        )
+
+        serialized_components = range_tl.component_manager.serialize_components()
+
+        range_tl.component_manager.clear()
+
+        range_tl.component_manager.deserialize_components(serialized_components)
```
FelipeDefensor, line 53: Also assert the properties were preserved

tests/timelines/range/test_range_timeline.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def test_overlap_validation(self, range_tl):
+        row_id = range_tl.rows[0].id
+        # Create initial range
+        c, _ = range_tl.create_component(
+            ComponentKind.RANGE, id=1, start=10, end=20, row_id=row_id
+        )
+        assert c is not None
+
+        # Try overlapping - SHOULD SUCCEED NOW
```
FelipeDefensor: Should SHOULD SUCCEED NOW be here?

tests/ui/cli/components/test_cli_components_add.py
```diff
@@ -2,6 +2,11 @@
 
 
 class TestAddBeat:
+    def test_wrong_timeline_kind_raises_error(self, cli, tls, tilia_errors):
+        cli.parse_and_run("timeline add wrong")
+
+        tilia_errors.assert_error()
+
```
FelipeDefensor: I think this was moved elsewhere (as it should) by one of the commits that this PR is stacked on top of.

tests/ui/timelines/harmony/test_harmony_ui.py
```diff
@@ -1,4 +1,4 @@
-from unittest.mock import patch, Mock
```
FelipeDefensor: Drop this change, unrelated to this PR.

tests/ui/timelines/hierarchy/test_hierarchy_timeline_ui.py
```diff
@@ -131,6 +131,27 @@ def test_add_post_end(self, tlui):
         assert tlui[0].get_data("post_end") != tlui[0].get_data("end")
         assert tlui[0].post_end_handle
 
+    def test_whisker_hover_changes_cursor(self, tlui):
+        # Regression: hovering the whisker grab tab didn't change the cursor
+        # because the VLine sits inside a QGraphicsItemGroup whose default
+        # routing prevented hover events from reaching the child.
+        from PySide6.QtCore import QEvent, QPointF, Qt
+        from PySide6.QtGui import QGuiApplication
+        from PySide6.QtWidgets import QGraphicsSceneHoverEvent
```
FelipeDefensor: Any reason for the local imports?

tilia/parsers/csv/range.py
```diff
@@ -0,0 +1,333 @@
...
+                    errors.append(
+                        f"start={start_measure}, end={end_measure} | "
+                        f"joined_with_next={raw!r} is not a valid boolean"
+                    )
+                    joined_flag = False
+
+            row = _get_or_create_row(timeline, row_name)
+
+            # Mirror the hierarchy parser's "pair start/end" logic so repeated
+            # measure numbers map to multiple distinct ranges.
```
FelipeDefensor, line 303: Is it worth it abstracting that behavior?

tests/ui/cli/timelines/test_cli_timelines_import.py
```diff
@@ -86,6 +87,72 @@ def test_beats(self, cli, beat_tl, tmp_path):
...
+
+        cli.parse_and_run(
+            f"timelines import range by-time --target-ordinal 1 "
+            f"--file {str(csv_path.resolve())}"
+        )
+
+        ranges = sorted(range_tl, key=lambda r: r.start)
+        assert ranges[0].get_data("label") == "first"
+        assert ranges[1].get_data("label") == "second"
+        assert {r.name for r in range_tl.rows} == {"A", "B"}
```
FelipeDefensor: Also assert range properties.

tests/ui/cli/timelines/test_cli_timelines_import.py
```diff
@@ -86,6 +87,72 @@ def test_beats(self, cli, beat_tl, tmp_path):
...
+        csv_path = tmp_csv(tmp_path, data)
+
+        cli.parse_and_run(
+            f"timelines import range by-measure --target-ordinal 1 "
+            f"--reference-tl-ordinal 2 --file {str(csv_path.resolve())}"
+        )
+
+        ranges = sorted(range_tl, key=lambda r: r.start)
+        assert ranges[0].get_data("label") == "first"
+        assert ranges[1].get_data("label") == "second"
```
FelipeDefensor: Also assert range properties.

tilia/timelines/range/components.py
```diff
@@ -0,0 +1,102 @@
...
+
+    @property
+    def end(self) -> float:
+        return self._end
+
+    @end.setter
+    def end(self, value: float) -> None:
+        prev_end = self._end
+        self._end = value
+        if self.post_end < value or self.post_end == prev_end:
```
FelipeDefensor, line 101: Add comments mentioning that it mirros start behavior.

tests/ui/cli/components/test_cli_components_add.py
FelipeDefensor: **Claude:** Addressed in f3f028ad.

tests/ui/timelines/harmony/test_harmony_ui.py
FelipeDefensor: **Claude:** Addressed in 4c6d2f52.

tests/ui/timelines/hierarchy/test_hierarchy_timeline_ui.py
FelipeDefensor: **Claude:** Addressed in 91afc12c (hoisted to module level).

tilia/timelines/range/timeline.py
```diff
@@ -0,0 +1,453 @@
...
+        self.scale = functools.partial(scale_segmentlike, self)
+        self.crop = functools.partial(crop_segmentlike, self)
+        self._restoring_state = False
+
+    def restore_state(self, prev_state: dict) -> None:
+        # Suppress join-cascade in delete_component while restoring: when
+        # a range's hash changes (e.g. post_end edit), restore_state deletes
+        # and recreates it. The cascade would otherwise clear an unrelated
+        # neighbor's joined_right, leaving the restored state inconsistent
+        # with the snapshot.
```
FelipeDefensor: Did not understand this. How does the hash relate to the join cascade? Explain this to me in the Claude Code session before adjusting.

tilia/timelines/range/timeline.py
```diff
@@ -0,0 +1,453 @@
...
+        if not row_id or row_id not in self.timeline.row_ids:
+            return False, "Invalid row ID."
+
+        if start >= end:
+            return False, "Start time must be before end time."
+
+        media_duration = get(Get.MEDIA_DURATION)
+        if start < 0 or end > media_duration:
+            return False, "Range is outside media bounds."
+
```
FelipeDefensor, line 69: We should also validate pre-start and post-end to see if they are in bounds.

tilia/timelines/range/timeline.py
```diff
@@ -0,0 +1,453 @@
...
+    def join(self, ranges: list[Range]) -> tuple[bool, str]:
+        if len(ranges) < 2:
+            return False, "Select at least two ranges."
+
+        row_id = ranges[0].row_id
+        if any(r.row_id != row_id for r in ranges):
+            return False, "Ranges must be in the same row."
+
+        ranges = sorted(ranges, key=lambda r: r.start)
```
FelipeDefensor: The Range class already has comparison operators that let us sort without specifying a key, right? That is what ORDERING_ATTRS is for. If that is the case correct other instances of this in the codebase.

tests/ui/cli/timelines/test_cli_timelines_import.py
FelipeDefensor: **Claude:** Addressed in bc2ee1f6.

tests/ui/cli/timelines/test_cli_timelines_import.py
FelipeDefensor: **Claude:** Addressed in bc2ee1f6 (same commit covers both by-time and by-measure cases).

tilia/timelines/range/components.py
FelipeDefensor: **Claude:** Addressed in 5305713d.

tilia/timelines/range/timeline.py
FelipeDefensor: **Claude:** Addressed in 73084a89 (validates against media bounds; coverage in test_pre_start_negative_rejected and test_post_end_past_duration_rejected).

tilia/timelines/range/timeline.py
FelipeDefensor: **Claude:** Confirmed — Range.ORDERING_ATTRS = ("start", "row_id") gives natural ordering (start primary, row_id tiebreaker). Replaced `sorted(.., key=lambda r: r.start)` with `sorted(..)` in production (1f448e53) and across the range test suite (ff987acb). Also covered `entries.sort(key=lambda e: e[0].start)` in parsers/csv/range.py — entries are (Range, bool) tuples, so `key=lambda e: e[0]` defers to Range.__lt__.

tilia/parsers/csv/range.py
FelipeDefensor: **Claude:** Marginal value, but probably yes. The two existing impls (this one and the hierarchy parser at tilia/parsers/csv/hierarchy.py:196-213) differ in two subtle ways:

1. The inner pop loop uses `>=` here vs `>` in hierarchy — the hierarchy version hangs forever if start==end (the outer loop checks `start < end`, the inner checks `start > end`, and the equal case satisfies neither so we never advance). That's effectively a bug in hierarchy.
2. Range needs the list of created components for join handling; hierarchy doesn't.

A shared helper `pair_start_end_times(start_times, end_times, on_pair)` (callback-based) would unify them and silently fix the hierarchy bug. Want me to do it in this PR or split it into a separate refactor commit on top of main? Leaving as-is for now until you confirm.

tilia/timelines/range/timeline.py
FelipeDefensor: **Claude:** Explanation:

"Hash" here = `TimelineComponent.hash` (tilia/timelines/base/component/base.py:46-47), a content hash over SERIALIZABLE attributes (start, end, post_end, joined_right, ...). `TimelineComponentManager.restore_state` (tilia/timelines/base/timeline.py:526-549) compares current vs snapshot hashes; any component whose hash *changed* is deleted and recreated.

"Join cascade" = the loop in `RangeTLComponentManager.delete_component` (timeline.py:224-233): when a range is deleted, walk every other range and clear its `joined_right` if it pointed at the deleted one.

The collision on undo:

1. A.joined_right == B.id.
2. User edits B's post_end. B's hash changes.
3. Undo. restore_state diff: B's *new* hash is in current but not snapshot → marked for delete; B's *old* hash is in snapshot but not current → marked for recreate.
4. During the delete phase, the cascade scans for c.joined_right == B.id, finds A, and clears A.joined_right.
5. B is recreated with the snapshot id. But A is *not* in the delete-or-recreate set (its hash didn't change), so it's never re-deserialized. A.joined_right stays None — diverging from the snapshot, which still had A→B.

Suppressing the cascade is correct because the snapshot already encodes the right join state for *every* range. A sibling that should have its joined_right cleared will be handled by re-deserializing it (if its hash changed) or left as-is (if it didn't). The cascade is only useful for *user-initiated* deletions, where there's no snapshot to fall back on.

Let me know how you'd like the comment phrased — happy to spell this out more in the code, or rework it differently if you don't like the flag-based approach.

tilia/parsers/csv/range.py
FelipeDefensor: I even think there is an issue for the hang you mentioned. I woud like the fix and refactor in a separate PR, with aproppriate tests.

FelipeDefensor: **Claude:** Acknowledged — leaving as-is in this PR. Will do the abstraction (and the hierarchy hang fix) in a separate PR with tests covering the start==end case for both timeline kinds.

tilia/timelines/range/timeline.py
FelipeDefensor: Now I get it. Briefly mentione the recreate due to hash change pattern in the comment, that should make it more understandable.

FelipeDefensor: **Claude:** Addressed in 0a52c32c — comment now leads with the hash-diff/recreate pattern, then explains why the cascade has to be suppressed during it.

tilia/timelines/range/timeline.py
```diff
@@ -0,0 +1,453 @@
...
+                if c.id != component.id and c.joined_right == component.id:
+                    self.timeline.set_component_data(c.id, "joined_right", None)
+        super().delete_component(component)
+
+
+def _validate_row_height(value: Any) -> bool:
+    return value is None or (isinstance(value, int) and value >= 10)
+
+
+VALID_LABEL_ALIGNMENTS = ("left", "center", "right")
```
FelipeDefensor: This is a frontend setting and therefore should not live in this backend module.

tilia/timelines/range/timeline.py
```diff
@@ -0,0 +1,453 @@
...
+    def default_color(self) -> str:
+        return settings.get("range_timeline", "default_range_color")
+
+    @property
+    def merge_separator(self) -> str:
+        return settings.get("range_timeline", "merge_separator")
+
+    @property
+    def is_empty(self) -> bool:
+        return len(self) == 0
```
FelipeDefensor: Is this already in Timeline?

tilia/ui/cli/timelines/add.py
```diff
@@ -42,18 +54,25 @@ def setup_parser(subparser):
         default=[4],
         help="Pattern as space-separated integers indicating beat count in a measure. Pattern will be repeated. Pattern '3 4', for instance, will alternate measures of 3 and 4 beats.",
     )
+    add_subp.add_argument(
+        "--row-height",
+        type=int,
+        default=None,
+        help="Per-timeline row height (range timelines only). Defaults to the global setting.",
+    )
```
FelipeDefensor, line 66: Do we have guards for passing --row-height as argument to other timeline types?

tilia/ui/cli/timelines/imp.py
```diff
@@ -23,37 +25,44 @@ def setup_parser(subparsers):
...
     )
     setup_import_file_and_target_args(parser)
 
 
 def setup_import_marker_and_hierarchy_parser(subparser):
     component_info = [
-        ("marker", "markers", ["mrk"]),
-        ("hierarchy", "hierarchies", ["hrc"]),
+        ("marker", "markers"),
+        ("hierarchy", "hierarchies"),
```
FelipeDefensor, line 52: Removing the aliases should be a separate commit.

tilia/ui/timelines/collection/collection.py
```diff
@@ -1451,19 +1453,20 @@ def get_timeline_ui_by_attr(self, attr: str, value: Any) -> TimelineUI | None:
...
+            (tlui for tlui in self._select_order if isinstance(tlui.timeline, kind)),
+            None,
         )
 
     def _get_choose_timeline_dialog(
         self,
         title: str,
         prompt: str,
-        kind: type(Timeline) | list[type(Timeline)] | None = None,
+        kind: type[Timeline] | list[type[Timeline]] | None = None,
```
FelipeDefensor, line 1572: These bracket changes should be a separate commit.

tilia/undo_manager.py
```diff
@@ -76,6 +76,7 @@ def redo(self):
         post(Post.APP_STATE_RESTORE, next_state)
 
         self.current_state_index += 1
+        post(Post.APP_STATE_UNDO_OR_REDO_DONE)
```
FelipeDefensor, line 79: Separate commit with message explaining rationale.

TESTING.md
```diff
@@ -43,6 +43,9 @@ def test_me(marker_tlui):
 ```
 You will find many examples of the former in the test suite, though. Refactors are welcome.
 
+## How to test code that depends on settings?
```
FelipeDefensor, line 66: Separate commit with message explaining rationale.

tilia/ui/cli/timelines/add.py
FelipeDefensor: **Claude:** No — argparse accepts `--row-height` for any `kind`, and `TLKIND_TO_KWARGS_NAMES` silently drops it for non-range kinds. `--beat-pattern` is set up the same way (accepted globally, only forwarded for `BeatTimeline`).

Want me to add a guard that errors when an arg is passed to a kind that doesn't take it (applied to both `--row-height` and `--beat-pattern`)? Or leave the silent-drop pattern intact?

tilia/ui/cli/timelines/imp.py
FelipeDefensor: **Claude:** Acknowledged — the alias removal is in the original `feat: range timeline` commit. I'll split it out into its own commit at final-squash time.

tilia/ui/timelines/collection/collection.py
FelipeDefensor: **Claude:** Acknowledged — the `type(Foo)` → `type[Foo]` modernization is in the original `feat: range timeline` commit. I'll split it out at final-squash time.

tilia/undo_manager.py
FelipeDefensor: **Claude:** Acknowledged — the `Post.APP_STATE_UNDO_OR_REDO_DONE` post is in the original `feat: range timeline` commit. I'll split it out at final-squash time with a message explaining the rationale (range UI redrawing after undo/redo).

TESTING.md
FelipeDefensor: **Claude:** Acknowledged — the TESTING.md addition is in the original `feat: range timeline` commit. I'll split it out at final-squash time with a message explaining the rationale (documenting the settings-test pattern uncovered while writing range tests).

tilia/timelines/range/timeline.py
FelipeDefensor: **Claude:** Addressed in 8a677f18 — moved `VALID_LABEL_ALIGNMENTS` to `tilia/ui/timelines/range/toolbar.py` (alongside the alignment buttons), updated the settings window's import to match.

tilia/timelines/range/timeline.py
FelipeDefensor: **Claude:** Addressed in 79e643e5 — yes, base `Timeline.is_empty` already exists with the same `len(self) == 0` implementation. Override dropped.

tilia/ui/cli/timelines/add.py
FelipeDefensor: Add the guard in this PR and create a PR for adding the guard for beat timelines.

tests/parsers/csv/test_range_from_csv.py
FelipeDefensor: **Claude:** Addressed in fb4d684f — direct-call helpers (`_import_by_time` / `_import_by_measure`) dropped; all tests now route through `_trigger_import_through_command`.

tests/parsers/csv/test_range_from_csv.py
FelipeDefensor: **Claude:** Addressed in 9d76f973 — switched from `sum(1 for r in ...)` to `len([...])`. (`len()` doesn't accept a generator, so a list comprehension is the closest spelling to your suggestion.)

tests/parsers/csv/test_range_from_csv.py
FelipeDefensor: **Claude:** Addressed in 1847b29d — `list(range_tl)[0]` replaced with `range_tl[0]` in `test_with_fractions`.

tests/parsers/csv/test_range_from_csv.py
FelipeDefensor: **Claude:** Addressed in 8fbb1e47 — added `RangeTimeline.get_ranges_by_row(row_id)` and applied it in tests/parsers/csv/test_range_from_csv.py and the `joined_right` cascade in tilia/timelines/range/timeline.py. Other call sites already had per-row context, so no further substitutions made sense.

tests/parsers/csv/test_range_from_csv.py
FelipeDefensor: **Claude:** Addressed in 76524211 — `test_flag_on_last_range_reports_error` now uses `range_tlui.timeline[-1].joined_right`.

tests/timelines/range/test_range_timeline.py
FelipeDefensor: **Claude:** Addressed in 305ffd29 — `test_serialize_components` now asserts `start`, `end`, and `row_id` for both r1 and r2.

tests/timelines/range/test_range_timeline.py
FelipeDefensor: **Claude:** Addressed in 9b4dc74e — `test_deserialize_components` now sorts the restored components and asserts `start`, `end`, and `row_id` match the originals.

tests/timelines/range/test_range_timeline.py
FelipeDefensor: **Claude:** Addressed in fefc2f69 — stale `# SHOULD SUCCEED NOW` marker removed from `test_overlap_validation`.

examples/range/by_measure_basic.csv
FelipeDefensor: **Claude:** Folder removed in 44c05930 (you authored the commit yourself). For the second bullet — I can file an issue in `TimeLineAnnotator/testing` if you want, but I need more detail than "not able to open files from the testing repo". Which files, what error, and which version? Happy to file once you share, or you can file directly.

tilia/ui/cli/timelines/add.py
FelipeDefensor: **Claude:** Addressed in b1348552 — `--row-height` now errors out via `tilia.errors.CLI_ADD_TIMELINE_ARG_NOT_APPLICABLE` when paired with a non-range kind. `--beat-pattern` will be handled in a separate PR off `main` as requested.

tilia/ui/timelines/range/context_menu.py
```diff
@@ -0,0 +1,189 @@
...
+
+        add_action("Rename row", self.on_rename_row)
+        add_action("Set row color", self.on_set_row_color)
+        if self.row.color is not None:
+            add_action("Reset row color", self.on_reset_row_color)
+        add_action("Set height for this row", self.on_set_row_height_for_row)
+        if self.row.height is not None:
+            add_action("Reset row height", self.on_reset_row_height_for_row)
+        row_idx = self.timeline_ui.timeline.row_index(self.row)
+        if row_idx is not None and row_idx > 0:
```
FelipeDefensor: Do you think we should defend against row index being <= 0 here? If so, we need at least a console warning.

FelipeDefensor: **Claude:** Good catch — analysis below, then what I changed.

`RangeTimeline.row_index` returns `int | None` (never negative). `self.row` was set from `get_row_by_y`, which only returns rows it found in `timeline.rows` or `None` (already early-returned at line 63). So in the well-formed case `row_idx ∈ {0, 1, …, row_count - 1}`. The two cases the old `row_idx is not None and row_idx > 0` guard collapsed:

- **`row_idx == 0`** — top row, no "Move row up". Intentional menu behavior, no warning warranted.
- **`row_idx is None`** — row found by y-coordinate but not in `timeline.rows` anymore. Stale UI reference, would be a bug.

Addressed in 5012c2c5 — only the `is None` case warns (via `tilia.log.logger`) and skips the move actions; the `== 0` branch stays silent. Let me know if you wanted the `== 0` case to warn too.

tilia/ui/timelines/range/drag.py
```diff
@@ -0,0 +1,274 @@
...
+        after_each=functools.partial(
+            after_each_drag, range_ui, extremity, partner_ui, partner_extremity
+        ),
+        on_release=functools.partial(on_drag_end, range_ui),
+    )
+
+
+def _start_frame_drag(range_ui: RangeUI, extremity: Extremity) -> None:
+    if extremity == "pre_start":
+        get_min_x = lambda: get(Get.LEFT_MARGIN_X)  # noqa: E731
```
FelipeDefensor: What is E731? Maybe add a comment on that?

tilia/ui/timelines/range/context_menu.py
FelipeDefensor: **Claude:** Switched from `logger.warning` to `logger.error` in 2fecd44c — matches the `collection.py:1095` precedent for developer-facing invariant violations and actually reaches the console (default threshold is ERROR; WARNING only surfaces under `dev.log_requests=True`).

tilia/ui/timelines/range/handles.py
```diff
@@ -0,0 +1,152 @@
...
+        def boundingRect(self) -> QRectF:
+            return self.shape().boundingRect()
+
+    class HLine(QGraphicsLineItem):
+        def __init__(self, x0: float, x1: float, y: float, width: int):
+            super().__init__()
+            self.set_position(x0, x1, y)
+            self.set_pen(width)
+            # Non-interactive: clicks should fall through to the timeline
+            # so rubber-band selection still works underneath.
```
FelipeDefensor: Not only rubber-band, but regular click selection too, right?

tilia/ui/timelines/range/toolbar.py
```diff
@@ -0,0 +1,212 @@
...
+    VALID_LABEL_ALIGNMENTS,
+    RangeTimeline,
+)
+from tilia.ui import commands
+from tilia.ui.timelines.toolbar import TimelineToolbar
+
+
+class _ToolbarSection(QWidget):
+    """A ribbon-style group: a row of action buttons with a section label
+    underneath. Mimics the Office toolbar grouping."""
```
FelipeDefensor: We can leave the Office reference out.

tilia/ui/timelines/range/drag.py
FelipeDefensor: **Claude:** E731 = "do not assign a lambda expression, use a def" (ruff/flake8 style rule). Addressed in 21b3ffaa by refactoring instead of explaining: lifted the per-extremity getters into nested `def`s with a ternary, matching the def-form already used in `start_body_drag` (drag.py:188-194). The four `# noqa: E731` suppressions are gone.

tilia/ui/timelines/range/handles.py
FelipeDefensor: **Claude:** Right — addressed in 95dba0f7. Comment now reads "click-to-select and rubber-band selection both still work on the ranges underneath".

tilia/ui/windows/inspect.py
```diff
@@ -257,13 +257,21 @@ def hide_or_show_field(self, field_name, value):
...
+            widget.blockSignals(True)
+            try:
+                if isinstance(widget, (QLineEdit, QLabel, QTextEdit)):
+                    widget.setText("")
+                elif isinstance(widget, QComboBox):
+                    widget.setCurrentIndex(0)
+                elif isinstance(widget, QSpinBox):
+                    widget.setValue(widget.minimum())
+            finally:
+                widget.blockSignals(False)
```
FelipeDefensor, line 292: Fix for the codebase as a whole. Should be a separate commit.

tilia/ui/timelines/range/toolbar.py
FelipeDefensor: **Claude:** Addressed in 7639c5bf — Office reference dropped from the `_ToolbarSection` docstring.

tilia/ui/windows/inspect.py
FelipeDefensor: **Claude:** Acknowledged — the `clear_widgets` block-signals change is in the original `feat: range timeline` commit. Same pattern as the four other deferred splits. I will split it out at final-squash time as a sibling to `51f2636c fix: block widget signals during programmatic value updates` (which already covered `set_widget_value`).

tests/ui/timelines/range/test_range_timeline_ui.py
```diff
@@ -0,0 +1,3744 @@
...
+        commands.execute("media.seek", 10)
+        commands.execute("timeline.range.add_range")
+        assert len(range_tlui) == 1
+        assert_range(
+            range_tlui,
+            0,
+            start=10,
+            end=10 + settings.get("range_timeline", "default_range_size"),
+        )
+
```
FelipeDefensor, line 177: We are missing a test_create_with_shortcut. Ensure that all the other commands that can be triggered with shortcuts have a specific test case for that.

tests/ui/timelines/range/test_range_timeline_ui.py
```diff
@@ -0,0 +1,3744 @@
...
+        file_path = tmp_path / "test.tla"
+
+        success = commands.execute("file.save_as", path=file_path)
+        assert success
+
+        commands.execute("file.new")
+        success = commands.execute("file.open", path=file_path)
+        assert success
```
FelipeDefensor: We have a helper for saving and reopening. Use it here and in similar places in this file, unless there is a good reason not to.

tests/ui/timelines/range/test_range_timeline_ui.py
```diff
@@ -0,0 +1,3744 @@
...
+        post(Post.TIMELINE_VIEW_LEFT_BUTTON_RELEASE)
+
+    def test_drag_joined_edge_after_file_load_moves_partner(
+        self, tilia, tluis, tilia_state, tmp_path
+    ):
+        # Regression: dragging the shared edge of joined ranges loaded from
+        # a file used to leave the partner behind because the partner lookup
+        # in drag.py used pre-save IDs that no longer matched the fresh IDs
+        # assigned at deserialize.
+        from tilia.requests import Post, post
```
FelipeDefensor: Any reason for local and not global import?

tests/ui/timelines/range/test_range_timeline_ui.py
```diff
@@ -0,0 +1,3744 @@
...
+
+        commands.execute("timelines.add.range", name="range")
+        range_tlui = tluis[0]
+
+        for _ in range(5):
```
FelipeDefensor, line 609: Add comment about advancing ID counter, if that is the purpose here.

tests/ui/timelines/range/test_range_timeline_ui.py
```diff
@@ -0,0 +1,3744 @@
...
+        # Media is 100s by default. Pasting a 10s-wide range at 95s would
+        # push the end past media duration → rejected by validator.
+        commands.execute("timeline.range.add_range", start=0, end=10)
+        click_range_ui(range_tlui[0])
+        commands.execute("timeline.component.copy")
+
+        click_timeline_ui(range_tlui, 400)
+        commands.execute("media.seek", 95)
+        commands.execute("timeline.component.paste")
+        assert len(range_tlui) == 1  # paste rejected, no new component
```
FelipeDefensor, line 1021: We should also test wether rejecting one out-of-bounds range still lets the other ranges be pasted normally.

tests/ui/timelines/range/test_range_timeline_ui.py
FelipeDefensor: **Claude:** Added `test_create_with_shortcut` in `cb0b0240`. As an audit pass for the other shortcut-driven range commands, I also added shortcut tests for `j` (join), `e` (merge), and `s` (split). 

The remaining shortcut-driven commands (`c` set_range_color, `Ctrl+Up` move_to_row_above, `Ctrl+Down` move_to_row_below) are registered with shortcuts but their actions aren't added to the toolbar — only to the right-click element context menu, which isn't permanently parented to a window. As a result, those shortcuts don't actually fire when the timeline is focused. Tests for those are deferred until the actions are wired into the toolbar (parallel to how hierarchy puts `increase_level` / `decrease_level` in both the toolbar and the context menu). Flag if you want me to do that fix in this PR or open a separate issue.

tests/ui/timelines/range/test_range_timeline_ui.py
FelipeDefensor: **Claude:** Refactored in `35af67bc` — replaced the inline `file.save_as` / `file.new` / `file.open` trio with `tests.utils.save_and_reopen(tmp_path)` across all save/load cases in the file (12 sites).

examples/range/by_measure_basic.csv
FelipeDefensor: The only files that worked are the ones the examples from range timelines we added. I get an invalid file error, or something like that. You could try opening them in TiLiA yourself to see if you can reproduce.

tests/ui/timelines/range/test_range_timeline_ui.py
FelipeDefensor: **Claude:** No reason — just leftover. Hoisted in `94817a24`. Same commit also hoists the other local imports in this file (`WindowKind`, `_row_at_y`, `RangeCommentsIcon`) for consistency.

tests/ui/timelines/range/test_range_timeline_ui.py
FelipeDefensor: **Claude:** Yes that is the purpose — added a comment in `94817a24`, mirroring the comment already on `test_save_load_preserves_join_with_advanced_id_counter`.

tests/ui/timelines/range/test_range_timeline_ui.py
FelipeDefensor: **Claude:** Added `test_paste_drops_only_out_of_bounds_member` in `4f2bcfb0` — copies a batch where one range is in bounds and one would be pushed past media duration, then asserts the in-bounds member still lands. Exercises the `if comp is None: continue` branch in `_create_pasted_ranges`.

examples/range/by_measure_basic.csv
FelipeDefensor: **Claude:** Reproduced. The .tla files in `TimeLineAnnotator/testing/tla/` are saved with a UTF-8 BOM, but `open_tla` in `tilia/file/file_manager.py` opens with `encoding="utf-8"` — which raises `json.JSONDecodeError: Unexpected UTF-8 BOM (decode using utf-8-sig)`, mapped to "Invalid TLA file". Verified locally on `test-marker-timeline.tla`.

The fix is one line: change `encoding="utf-8"` → `encoding="utf-8-sig"` in `open_tla` (and likely the metadata-import path at line 204 for parity). The bug is pre-existing on `main`, unrelated to range. Want me to open a separate PR off `main`?

FelipeDefensor: Everything should be already saved as utf-8. How can fix that and commit directly to main on the testing repo.

tests/ui/timelines/range/test_range_timeline_ui.py
FelipeDefensor: I don't want them in the toolbar to reduce clutter. I think having the option to drag to another row and the context menu is enough. I tested, and those shortcuts do not work in the live app. There should be a convenient, canonical way provided by Qt to make them work without having to add the action to the toolbar. You can create a PR in this branch to address this and report in this thread.

FelipeDefensor: Also create a PR to drop the c shortcut, it conflicts semantically with c for create child in hierarchy timelines.

FelipeDefensor: **Claude:** Addressed in two commits:

- `1367a23c fix: wire range Ctrl+Up/Down via per-view QShortcut` — for context-menu-only commands the QAction has no persistent parent widget, so Qt's shortcut routing never sees the key sequence. Bind a per-view `QShortcut` in `RangeTimelineUI.__init__` so the keystroke is reachable without polluting the toolbar. The `selector=FIRST` + `@with_elements` gate already in the callback makes the binding a no-op unless a range timeline has selected elements, so it stays harmless when the same keys belong to another kind (hierarchy uses Ctrl+Up/Down for level changes). Coverage: `test_move_to_row_above_with_shortcut` / `test_move_to_row_below_with_shortcut`.

- `95831cb6 fix: drop 'c' shortcut from range set_range_color` — the keybinding goes away (semantic conflict with hierarchy's `c` for create_child); the action stays in the right-click context menu.

examples/range/by_measure_basic.csv
FelipeDefensor: **Claude:** Audited the testing repo: 28 of 36 .tla fixtures had a UTF-8 BOM, 8 didn't (so this matches the user-visible symptom — the ones that *did* open were the 8 without). I cloned `TimeLineAnnotator/testing`, stripped the BOM from those 28, validated each as utf-8 JSON, and committed locally (`58448f2 Strip UTF-8 BOM from .tla files`). The push to main was blocked by the sandbox as a shared-systems action — let me know if you want me to retry once it's whitelisted, or if you'd rather pull the commit yourself: it's in `C:/Users/Felipe/dev/tmp/tilia-testing` on `main`.

tilia/ui/timelines/range/timeline.py
```diff
@@ -0,0 +1,988 @@
...
+        self.update_height()
+        self.update_row_labels()
+
+        listen(self, Post.SETTINGS_UPDATED, self.on_settings_updated)
+        listen(self, Post.RANGE_TIMELINE_CREATED, self._delete_row_highlight)
+        listen(self, Post.RANGE_TIMELINE_CLICKED, self._delete_row_highlight)
+
+        post(Post.RANGE_TIMELINE_CREATED)
+        # Highlight is recreated after the created event so existing
+        # highlights on other range timelines have already been removed.
```
FelipeDefensor: Confirm that this is really the reason.

tilia/ui/timelines/range/timeline.py
```diff
@@ -0,0 +1,988 @@
...
+
+    def get_data(self, attr: str) -> Any:
+        if attr == "height":
+            return self.height
+        else:
+            return super().get_data(attr)
+
+    @property
+    def row_height(self) -> int:
+        return self.timeline.row_height or settings.get(
```
FelipeDefensor: Presupposes that the height can't be zero. Confirm that we are preventing the user from setting that.

tilia/ui/timelines/range/timeline.py
```diff
@@ -0,0 +1,988 @@
...
+            "range_timeline", "default_row_height"
+        )
+
+    @property
+    def label_alignment(self) -> str:
+        return settings.get("range_timeline", "default_label_alignment")
+
+    @property
+    def height(self) -> int:
+        return int(sum(self.row_height_for(r) for r in self.rows)) + 20
```
FelipeDefensor: Extract the 20 into a setting called range timeline margin or something.

tilia/ui/timelines/range/timeline.py
```diff
@@ -0,0 +1,988 @@
...
+            self.update_highlight_position()
+        for element in self:
+            element.update_position()
+        self.update_row_labels()
+        self.update_height()
+
+    def update_row_labels(self) -> None:
+        current_row_ids = [row.id for row in self.timeline.rows]
+
+        for row_id in list(self.row_labels.keys()):
```
FelipeDefensor, line 300: Add short somments explaining what each of the loops do.

tilia/ui/timelines/range/timeline.py
```diff
@@ -0,0 +1,988 @@
...
+            self._apply_label_font(label, self.row_height_for(row))
+
+        for row_id, label in self.row_labels.items():
+            label.set_position(
+                get(Get.LEFT_MARGIN_X) - 2, self.row_y(self.row_index(row_id)) - 2
+            )
+
+    def _apply_label_font(self, label: LevelLabel, row_height: int) -> None:
+        # Shrinks the row label so it fits when the user picks a small row
+        # height. Without this clamp the default Arial 9pt overflows into
```
FelipeDefensor: Don't mention the exact default as the comment might get stale in the future.

tilia/ui/timelines/range/timeline.py
```diff
@@ -0,0 +1,988 @@
...
+            ),
+        )
+        font = QFont(label.font())
+        font.setPixelSize(target)
+        label.setFont(font)
+
+    def _get_row_argument(
+        self, row: RangeTimeline.Row | None = None
+    ) -> RangeTimeline.Row | None:
+        return row or self.selected_row
```
FelipeDefensor: What is this for? Maybe use a more descriptive name and/or a docstring? I am not sure.

tilia/ui/timelines/range/timeline.py
```diff
@@ -0,0 +1,988 @@
...
+        return component is not None
+
+    @with_elements
+    def on_join_ranges(self, elements: list[RangeUI]) -> bool:
+        success, reason = self.timeline.component_manager.join(
+            self.elements_to_components(elements)
+        )
+        if not success:
+            post(Post.DISPLAY_ERROR, "Error joining ranges", reason)
+        return success
```
FelipeDefensor, line 532: We have registered this command with cls.register_timeline_command, which means that it will be wrapped in TImelineUIs.on_timeline_command, which already has error handling. Confirm if that is the case, and fix if necessary. Same goes for the other commands below.

tilia/ui/timelines/range/timeline.py
```diff
@@ -0,0 +1,988 @@
...
+            success, height = get(
+                Get.FROM_USER_INT,
+                "Set default row height",
+                "Insert new default row height",
+                value=self.row_height,
+                minValue=10,
+            )
+            if not success:
+                return False
+        return get(Get.TIMELINE_COLLECTION).set_timeline_data(
```
FelipeDefensor, line 691: Comment on why we are calling TimelineUIs instead of self.timeline here?

tilia/ui/timelines/range/timeline.py
```diff
@@ -0,0 +1,988 @@
...
+            return False
+        idx = self.timeline.row_index(row)
+        if idx is None or idx == 0:
+            return False
+        return self.timeline.reorder_row(row, idx - 1)
+
+    def on_move_row_down(self, row: RangeTimeline.Row | None = None) -> bool:
+        row = self._get_row_argument(row)
+        if row is None:
+            return False
```
FelipeDefensor: Is it worth it to refactor this into a @with_rows decorator to mirror the @with_elements behavior? If so, do it.

tilia/ui/timelines/range/timeline.py
```diff
@@ -0,0 +1,988 @@
...
+        changed = False
+        for e in elements:
+            if e.get_data("post_end") != e.get_data("end"):
+                self.timeline.set_component_data(e.id, "post_end", e.get_data("end"))
+                changed = True
+        return changed
+
+    def row_y(self, row_index: int) -> float:
+        # Cumulative top-y of the row at `row_index`. Each row may have
+        # an explicit `height`; rows that don't fall back to the timeline's
```
FelipeDefensor: I understand it, but the second sentence reads a bit weird.

tilia/ui/timelines/range/timeline.py
```diff
@@ -0,0 +1,988 @@
...
+        if row_index <= 0:
+            return 0.0
+        total = 0.0
+        for row in self.rows[:row_index]:
+            total += self.row_height_for(row)
+        return total
+
+    def row_height_for(self, row: RangeTimeline.Row | None) -> int:
+        if row is None or row.height is None:
+            return self.row_height
```
FelipeDefensor: A bit late to comment on this, but would naming self.row_height self.default_row_height be less misleading?

tilia/ui/timelines/range/timeline.py
```diff
@@ -0,0 +1,988 @@
...
+
+    def update_height(self) -> None:
+        super().update_height()
+        self.update_highlight_position()
+
+    def on_horizontal_arrow_press(self, arrow: str) -> None:
+        if not self.has_selected_elements:
+            return
+        if arrow not in ("right", "left"):
+            raise ValueError(f"Invalid arrow '{arrow}'.")
```
FelipeDefensor: Let's not raise here to prevent crashes in production. Display an error instead.

tilia/ui/timelines/range/timeline.py
```diff
@@ -0,0 +1,988 @@
...
+
+        if arrow == "right":
+            self._deselect_all_but_last()
+        else:
+            self._deselect_all_but_first()
+
+        selected = self.element_manager.get_selected_elements()[0]
+        same_row = sorted(
+            (e for e in self if e.get_data("row_id") == selected.get_data("row_id")),
+            key=lambda e: e.get_data("start"),
```
FelipeDefensor: RangeUI already implements its own sorting logic through Range.ORDERING_ATTRS.

tilia/ui/timelines/range/timeline.py
```diff
@@ -0,0 +1,988 @@
...
+            if idx == 0:
+                return
+            target = same_row[idx - 1]
+
+        self.deselect_element(selected)
+        self.select_element(target)
+
+    def on_vertical_arrow_press(self, arrow: str) -> None:
+        if arrow not in ("up", "down"):
+            raise ValueError(f"Invalid arrow '{arrow}'.")
```
FelipeDefensor: Same as above.

tilia/ui/timelines/range/timeline.py
FelipeDefensor: **Claude:** Confirmed and rewrote the comment in 8a78ea80. The post fires every `_delete_row_highlight` listener — including this timeline's, since we registered just above — which collapses the "only the active timeline shows the highlight" invariant down to "no timeline shows it". The recreate after the post puts ours back, leaving this one as the active timeline. The new comment spells that out instead of only mentioning other timelines.

tilia/ui/timelines/range/timeline.py
FelipeDefensor: **Claude:** Addressed in 6b4be5ba — extracted the literal into `range_timeline.bottom_margin = 20` alongside the other range-timeline settings. `RangeTimelineUI.height` now reads it from `settings.get(...)`.

tilia/ui/timelines/range/timeline.py
FelipeDefensor: **Claude:** Addressed in 0ef54712 — one-line note on each of the three loops (drop-removed / create-and-refresh / reposition).

tilia/ui/timelines/range/timeline.py
FelipeDefensor: **Claude:** Addressed in 227d6c35 — `default Arial 9pt` replaced with `default font size` so the comment doesn't pin to whatever default Qt happens to use today.

tilia/ui/timelines/range/timeline.py
FelipeDefensor: **Claude:** Addressed in 6e5da189 — renamed to `_resolve_row` and added a docstring spelling out the explicit-or-selected fallback (the optional `row` argument lets context-menu callbacks pass the clicked row directly while keyboard-driven callers fall through to `selected_row`).

tilia/ui/timelines/range/timeline.py
FelipeDefensor: **Claude:** Confirmed at the timeline-default level. `_validate_row_height` (`tilia/timelines/range/timeline.py:246-247`) requires `value >= 10` (or `None`) and is wired into `validators["row_height"]`, so `set_data("row_height", 0)` fails through `validate_set_data` → `(None, False)` and the value is never stored. The GUI dialog uses `minValue=10` as a UX guard, and the CLI `--height` flag goes through the same `set_timeline_data` path. The `or` on line 144 is then defensive against legacy/corrupt files that might still carry an explicit `0`.

One small gap I noticed while looking: `RangeTimeline.set_row_height(row, height)` (per-row override) and `Row.from_dict` don't share that validator, so `row.height` could theoretically land at `0` from a hand-edited save. Want me to extend the guard to those paths in a follow-up commit?

tilia/ui/timelines/range/timeline.py
FelipeDefensor: **Claude:** Investigated — the manual `post(Post.DISPLAY_ERROR, ...)` blocks aren't redundant with what `TimelineUIs.on_timeline_command` provides. That wrapper (`collection.py:258-315`) handles two cases:

1. The callback raises an uncaught exception → shows `tilia.errors.COMMAND_FAILED` and rolls state back via `Post.APP_STATE_RECOVER`.
2. The callback returns the wrong type → shows `tilia.errors.INVALID_RETURN_VALUES_FOR_COMMAND`.

What it does *not* do is surface the user-facing reason from a `(False, "reason")` return value. When `result == False`, it just skips the `APP_STATE_RECORD` post and moves on silently. So the manual `post(Post.DISPLAY_ERROR, "Error joining ranges", reason)` here (and in `merge`/`split`/`separate`) is what actually tells the user *why* the operation failed.

A cleaner refactor would teach `on_timeline_command` to handle `(success, reason)` returns from the callback so this pattern could be lifted out of every command. That's a cross-kind change though; happy to do it as a follow-up if you want, but for this PR the manual handlers are correct as-is.

tests/ui/timelines/range/test_range_timeline_ui.py
FelipeDefensor: **Claude:** Reworked the Ctrl+Up/Down wiring in 63609710 — the per-view QShortcut from the previous attempt is gone. The new approach piggybacks on the existing `Post.TIMELINE_KEY_PRESS_*` flow that already powers arrow navigation: `TimelineView.keyPressEvent` now also detects `Ctrl+Up`/`Ctrl+Down` and posts `TIMELINE_KEY_PRESS_CTRL_UP`/`CTRL_DOWN`. `TimelineUIs.on_ctrl_arrow_press` dispatches once per kind to each timeline's `on_ctrl_vertical_arrow_press`, which in turn calls `commands.execute(...)` so the command goes through `on_timeline_command` (undo recording, error handling) and `@with_elements` keeps it a no-op without selection. Added keypress tests for both range (move-to-row-above/below) and hierarchy (increase/decrease level) to confirm the routing.

The `c` shortcut drop landed as a separate commit (41be646f, byte-identical to your 95831cb6 — the rebase that removed the QShortcut commit re-applied yours under a new SHA). It's already on the range branch. Since the file it touches doesn't exist on `main`/`dev` yet, a standalone PR isn't viable until the range PR merges; happy to open a follow-up off `dev` after merge if you'd still like that tracked separately.

tilia/ui/timelines/range/timeline.py
FelipeDefensor: No guards against hand-edited saves, we provide no support for them.

tilia/ui/timelines/range/timeline.py
FelipeDefensor: Perfect. Let's not touch it now, instead open an issue mentioning the potential refactor.

tests/parsers/csv/test_range_from_csv.py
```diff
@@ -250,68 +250,68 @@ def test_join_only_within_same_row(self, range_tl):
...
-    def test_gap_between_joined_ranges_reports_error(self, range_tl):
+    def test_gap_between_joined_ranges_reports_error(self, range_tlui):
         data = "start,end,row,joined_with_next\n0,10,A,true\n20,30,A,false"
-        success, errors = _import_by_time(range_tl, data)
+        success, errors = _trigger_import_through_command("time", data)
         assert success
         assert any("must equal" in e for e in errors)
-        ranges = sorted(range_tl, key=lambda r: r.start)
+        ranges = sorted(range_tlui.timeline, key=lambda r: r.start)
         assert ranges[0].joined_right is None
```
FelipeDefensor: Could we do range_tlui[0].get_data("joined_right") is None instead? If so, refactor every occurrence.

tilia/timelines/range/timeline.py
```diff
@@ -368,6 +366,12 @@ def get_row_by_index(self, index: int) -> Row | None:
             return None
         return self.rows[index]
 
+    def get_ranges_by_row(self, row_id: str) -> list[Range]:
+        """Return ranges on the given row, ordered by start time."""
+        return sorted(
+            (r for r in self if r.row_id == row_id), key=lambda r: r.start
```
FelipeDefensor: Ranges know how to sort themselves, we do not need to pass a key here.

tilia/ui/timelines/range/timeline.py
FelipeDefensor: **Claude:** Filed as #510. Left this PR untouched.

tilia/ui/timelines/range/timeline.py
FelipeDefensor: **Claude:** Acknowledged — leaving the per-row paths as-is.

tilia/ui/timelines/range/timeline.py
FelipeDefensor: **Claude:** Addressed in 74739494 — added a comment explaining that going through `TIMELINE_COLLECTION.set_timeline_data` (rather than `self.timeline.set_data`) is what posts `Post.TIMELINE_SET_DATA_DONE`, which in turn refreshes the UI when the default row height changes.

tilia/ui/timelines/range/timeline.py
FelipeDefensor: **Claude:** Worth it. Added `@with_row` in d7e78272, mirroring `@with_elements` — fall back to `self.selected_row` when no `row` is passed, return `False` if neither is available, otherwise call the body with a guaranteed-non-None row.

Applied to every method that previously rolled the same fallback by hand: `on_remove_row`, `on_rename_row`, `on_set_row_color`, `on_reset_row_color`, `on_set_row_height_for_row`, `on_reset_row_height_for_row`, `on_move_row_up`, `on_move_row_down`, and `on_add_range`. `_resolve_row` is gone (its only remaining caller, `on_split_range`, has its own conditional fallback because the "split all rows" toggle deliberately bypasses any row resolution). `on_reorder_row` stayed undecorated — reordering needs an explicit target with no selected-row fallback, so the decorator's behavior would be wrong there.

tilia/ui/timelines/range/timeline.py
FelipeDefensor: **Claude:** Reworded in 24c4d6d5 — now reads "Each row may set its own \`height\`; rows without one inherit the timeline's default row height."

tilia/ui/timelines/range/timeline.py
FelipeDefensor: **Claude:** Yes, less misleading. Renamed `RangeTimelineUI.row_height` → `default_row_height` in 926cc63b, and updated the call sites + tests. Backend `RangeTimeline.row_height` is unchanged so the SERIALIZABLE format stays compatible with existing files; the rename is UI-side only.

tilia/ui/timelines/range/timeline.py
FelipeDefensor: **Claude:** Addressed in e13a6411 — replaced the raise with `logger.error(...)` + early return, matching the precedent set in 2fecd44c (range/context_menu.py) for developer-facing invariant violations. The default console threshold is ERROR, so the diagnostic still surfaces; production no longer crashes on a bad arrow value.

tilia/ui/timelines/range/timeline.py
FelipeDefensor: **Claude:** Addressed in e10ba0cd — dropped the explicit `key=lambda e: e.get_data("start")`. `TimelineUIElement.__lt__` compares by `get_data("ordinal")`, which routes through `Range.ordinal` → the tuple of `Range.ORDERING_ATTRS = ("start", "row_id")`. Since the same-row filter pins `row_id`, the natural ordering already orders by `start`.

tilia/ui/timelines/range/timeline.py
FelipeDefensor: **Claude:** Same fix as line 904 — both raises went away in e13a6411 (logger.error + return).

tilia/ui/timelines/range/timeline.py
FelipeDefensor: This is unreleased, so you can break existing files. Rename the backend. Also update the files in the TimeLineAnnotator/testing repo.

tilia/ui/timelines/hierarchy/timeline.py
```diff
@@ -57,14 +57,14 @@ def register_commands(cls, collection: TimelineUIs):
             (
                 "decrease_level",
                 "Move down a level",
-                "Ctrl+Down",
+                "",
```
FelipeDefensor, line 66: Add a comment explaining that this is wired in a different way.

FelipeDefensor: Do the same in analogous places here and for range timelines.

tilia/timelines/range/timeline.py
FelipeDefensor: **Claude:** Addressed in 42241e80 — dropped the explicit `key=lambda r: r.start` from `get_ranges_by_row`. `Range.ORDERING_ATTRS = ("start", "row_id")` already drives `__lt__`, so `sorted(...)` returns the same order without the key.

tests/parsers/csv/test_range_from_csv.py
FelipeDefensor: **Claude:** Addressed in 0b759998 — every `joined_right` access is now `range_tlui[N].get_data("joined_right")`. UI elements are kept sorted by `(start, row_id)` ordinal, so the `sorted(range_tlui.timeline)` calls drop out and indexing is enough.

Same pass also folds in `len(range_tlui.timeline)` → `len(range_tlui)` and the property accesses in `test_basic` / `test_propagates_color_and_comments` for consistency. Backend access is kept where the API only exists on the backend timeline (`.rows`).

tilia/ui/timelines/range/timeline.py
FelipeDefensor: **Claude:** Addressed in b4c15226 — backend renamed to `RangeTimeline.default_row_height` (SERIALIZABLE list, validators key, `__init__` kwarg, plus the `update_row_height` -> `update_default_row_height` updater so `UPDATE_TRIGGERS` resolves). The CLI flag stays `--row-height` (user-facing, terse) but argparse `dest="default_row_height"` routes to the renamed kwarg.

For the testing repo: I checked TimeLineAnnotator/testing/tla — none of the existing .tla fixtures contain a range timeline, so there's nothing to migrate there. New range fixtures saved off this branch will use the new key automatically.

tilia/ui/timelines/hierarchy/timeline.py
FelipeDefensor: **Claude:** Addressed in 2ee4586d — added a comment above the `decrease_level` / `increase_level` registrations explaining that Ctrl+Down / Ctrl+Up are dispatched through `Post.TIMELINE_KEY_PRESS_CTRL_UP/DOWN` (see `TimelineView` and `TimelineUIs.on_ctrl_arrow_press`), so the shortcut field is intentionally empty — attaching the chord to the `QAction` as well would produce a Qt "Ambiguous shortcut" warning.

FelipeDefensor: **Claude:** Addressed in 2ee4586d — same comment now sits above the range timeline pair (`move_to_row_above` / `move_to_row_below`, the analogous Ctrl+Up / Ctrl+Down commands). `move_row_up` / `move_row_down` are not wired through this path (they're row-reordering commands, not range-navigation), so they didn't need a note.

azfoo (approved):
## Correctness fixes
- **Scale/crop bug**: resizing media duration or cropping silently corrupted a range's `pre_start`/`post_end` (shared code, also fixed the same bug for Hierarchy).
- **Copy/paste**: pasting dropped the join link on the anchor element, and never preserved `pre_start`/`post_end` at all.
- **Split**: left a stale `post_end` on the left half when the original range had a post-extension; also removed a redundant boundary-nudging special case (split at an existing join boundary is now a clean no-op — Separate is the only way to undo a join).
- **Merge**: now keeps matching label/comments instead of duplicating them.
- **Row color validation**: `add_row`/`set_row_color` now validate and surface a proper error instead of silently accepting garbage.
- **CLI**: invalid row height now surfaces an error instead of silently no-opping; import subcommands (`tl_type`, `measure_or_time`) now require an argument instead of falling through with `None`.
- **Zoom/drag**: a floating-point edge case in body/whisker-drag math could leave `pre_start`/`post_end` a hair off 0 or `media_duration` — now snapped to the exact boundary (same fix ported to the general drag code all timelines share).

## Performance
- Join-chain handle-visibility updates were O(n³); now O(n²) via a precomputed reverse index.

## Toolbar redesign
- Cut from 13 widgets to 5: Ranges group (Add range + Join/Merge dropdown + Separate/Split dropdown), Rows group (Add row above/below dropdown), and a Split-mode toggle.
- Remove-row dropped off the toolbar entirely (context-menu-only now); align-labels/toggle-extensions dropped too (both are Settings-only, avoiding duplicated state).
- Dropdowns remember the last-used action (click the main button to repeat it; the arrow opens the full menu) and now have a visibly distinct divider/arrow so the split-button affordance is obvious.
- Section labels ("Ranges"/"Rows") removed now that the groups are small enough to read at a glance.

## Shared-shortcut handling
- Detected and fixed a same-kind shortcut collision gap in the shared-dispatch system.
- Fixed `e`/`s` (merge/split, shared with Hierarchy) silently losing their menu display hint when Qt strips the `QAction` shortcut to avoid an ambiguity warning — now shown as a `\tKEY` text hint with no change to actual dispatch behavior.

## Stability
- Found and fixed a genuine `STATUS_HEAP_CORRUPTION` crash (stale `.cleanup()` call surviving a cursor-handling migration, affected both Range and Hierarchy).
- Hardened `commands.reset()` against leaving live `QShortcut` objects dangling.

## Known, deliberately left alone
- Row default-naming can produce a duplicate name after deleting a middle row and adding a new one — documented with a test, not fixed.
- Range's merge doesn't cross-check color the way Hierarchy does (keep-if-unanimous-else-default) — Range has no meaningful per-position default to fall back to, so the survivor just keeps its own color. Left as-is, deliberate.

azfoo:
to decide: should all buttons have shortcuts? separate currently does not. would key chords help give more options? eg. join [J] -> separate [Shift+J]?

FelipeDefensor:
> to decide: should all buttons have shortcuts? separate currently does not. would key chords help give more options? eg. join [J] -> separate [Shift+J]?

I don't think they should necessarily have, by default. Eventually, it would be nice to have customisable keyboard shortcuts where users could assign shortcuts to ones that don't.

PR #517, opened 2026-05-07
fix: release DragManager when timeline element is deleted (#515)

### LINKED ISSUE #515: Marker deleted mid-drag leaves orphan DragManager listener -> KeyError on next drag
## Steps to reproduce

1. Click and hold on a marker to start dragging it.
2. While still dragging (mouse button held down), press Delete to delete the marker.
3. Release the mouse, then start a new drag anywhere in the timeline.

The next drag throws `KeyError: '<deleted-marker-id>'`.

## Root cause

`MarkerUI.on_left_click` (`tilia/ui/timelines/marker/element.py:113-114`) calls `setup_drag()`, which creates a `DragManager` that subscribes to `Post.TIMELINE_VIEW_LEFT_BUTTON_DRAG` / `Post.TIMELINE_VIEW_LEFT_BUTTON_RELEASE`. The subscription is only torn down inside `DragManager.on_mouse_release` (`tilia/ui/timelines/drag.py:40-43`).

If the marker is deleted before the mouse-release fires, the marker element is gone but its `DragManager` listener stays in `_posts_to_listeners[TIMELINE_VIEW_LEFT_BUTTON_DRAG]`. The next drag dispatches to the stale listener, which calls `after_each_drag(drag_x)` -> `self.set_data("time", time_x_converter.get_time_by_x(drag_x))`. `set_data` tries to look up the (now phantom) component by id and raises `KeyError`.

## Fix sketch

When a marker element is deleted, also call `self.drag_manager.on_release()` (or otherwise stop_listening) if `self.drag_manager` is set. The existing `on_double_left_click` already does this defensively (`element.py:120-122`); the deletion path needs the same guard.

## Discovery

Found by exploration through the MCP server (PR #3 in fork) — clicking a marker via `Post.TIMELINE_VIEW_LEFT_CLICK`, deleting it via `timeline.component.delete`, then dragging anywhere reliably triggered the KeyError.

FelipeDefensor:
Closes #515.

## Summary

- Clicking a timeline element arms a `DragManager` that subscribes to `TIMELINE_VIEW_LEFT_BUTTON_DRAG` / `_RELEASE`. The subscription was only torn down by `DragManager.on_mouse_release`, so deleting the element before mouse release left a stale listener whose `after_each` callback (e.g. `MarkerUI.after_each_drag`) referred to a now-phantom component. The next drag anywhere in the timeline re-entered that callback on the deleted element and crashed with `KeyError` on `id_to_component`.
- Fix is in the base `TimelineUIElement.delete`: if a `drag_manager` is attached, unsubscribe it before tearing down the rest of the element. The same orphan path exists for hierarchy, beat and harmony elements (all use the same base `delete` and the same `drag_manager` field), so they get the fix for free.
- Adds `DragManager.cleanup` for the unsubscribe-only case, distinct from `on_mouse_release`. The latter still fires the user's `on_release` callback (the `on_drag_end` state-record `post`); the deletion path must not, since the `Delete` command is itself the recorded user action.

## Why a base-class fix

The reporter's sketch in #515 is marker-only, but the bug is structural: every element type that calls `setup_drag()` from `on_left_click` shares the same listener-leak window. Putting the cleanup in `TimelineUIElement.delete` (guarded by `getattr(self, "drag_manager", None)`) covers all current and future kinds without each element having to remember.

## Test plan

- New regression test `TestDrag::test_delete_mid_drag_does_not_leak_listener` in `tests/ui/timelines/marker/test_marker_timeline_ui.py` reproduces the original repro: click a marker (arms `drag_manager`), delete it via `timeline.component.delete`, then post a `LEFT_BUTTON_DRAG`. Without the fix this raises `KeyError: '<marker-id>'`; with it, the post is silent.
- Full test suite passes locally (1181 passed, 10 skipped, 2 xfailed, 2 xpassed).

## Out of scope

Several `on_double_left_click` handlers (`marker`, `beat`, `harmony.harmony`, `harmony.mode`, `hierarchy`) call `self.drag_manager.on_release()` to "cancel" a single-click-armed drag. That call only fires the user callback — it does **not** unsubscribe the listener. In normal flow the `LEFT_BUTTON_RELEASE` between the two clicks already cleans up, so it isn't observable, but the call is misleading. Left as-is to keep this PR focused on #515.

### DIFF
--- a/tests/ui/timelines/marker/test_marker_timeline_ui.py
+++ b/tests/ui/timelines/marker/test_marker_timeline_ui.py
@@ -311,8 +311,6 @@ class TestDrag:
         click_marker_ui(marker_tlui[0])
         assert marker_tlui[0].drag_manager is not None  # arm precondition
 
-        commands.execute("timeline.component.delete")
-
         # Without the fix, this dispatches into the deleted element's
         # DragManager and raises KeyError on set_data("time", ...).
         post(
@@ -320,7 +318,11 @@ class TestDrag:
             int(time_x_converter.get_x_by_time(20)),
             0,
         )
+        commands.execute("timeline.component.delete")
+
         post(Post.TIMELINE_VIEW_LEFT_BUTTON_RELEASE)
+
+        assert marker_tlui.is_empty
 
 
 class TestElementContextMenu:
--- a/tests/ui/timelines/marker/test_marker_timeline_ui.py
+++ b/tests/ui/timelines/marker/test_marker_timeline_ui.py
@@ -556,7 +556,10 @@ class TestInspect:
 
         assert marker_tlui[0].get_data("comments") == ""
 
-    def test_set_attribute_with_multiple_selected(self, marker_tlui, tluis):
+    def test_set_attribute_with_multiple_selected(
+        self, marker_tlui, tluis, tilia_state
+    ):
+        tilia_state.current_time = 0
         commands.execute("timeline.marker.add")
         commands.execute("media.seek", 10)
         commands.execute("timeline.marker.add")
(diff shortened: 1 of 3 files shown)


azfoo (approved):

azfoo:
@FelipeDefensor you might want to double check this?

PR #520, opened 2026-05-08
fix: detect media_metadata changes for save-changes prompt (#377)

### LINKED ISSUE #377: Save changes dialog does not appear when modifying only "notes" metadata field
Which means the user loses changes unless they were saved explicitly before closing. I haven't tested this manually for other metadata fields, but we do have automatic tests for those, if I remember correctly.

FelipeDefensor:
Closes #377.

## Summary

\`FileManager.is_file_modified\` compared the current app state against \`self.file.__dict__\`, but for \`media_metadata\` **both sides referenced the same live \`MediaMetadata\` object**: \`on_set_media_metadata_field\` mutates \`self.file.media_metadata\` in place, and \`get_app_state\` takes \`dict(self.file.media_metadata)\` of that same mutated object. The comparison was always equal, so editing notes never set the file to "modified" and the save-changes dialog never appeared on close.

The reporter flagged "notes" specifically, but it's not unique to that field — the title, custom fields, etc. were broken in the same way. Verified manually before the fix:

\`\`\`
Initial is_file_modified: False
After title change is_file_modified: False
After notes change is_file_modified: False
\`\`\`

After the fix:

\`\`\`
Initial is_file_modified: False
After title change is_file_modified: True
After notes change is_file_modified: True
\`\`\`

## Approach

Track a snapshot of \`media_metadata\` taken at the last save / file-load.

- \`self._saved_metadata\` is set by a property setter on \`file\`, so every place that reassigns the file (\`__init__\`, \`_setup_file\`, \`save\`, \`update_file\`, \`app.on_open\`'s \`self.file_manager.file = file\`) refreshes the snapshot for free — no need to thread an extra method-call through every call site.
- \`is_file_modified\` compares \`dict(current_data[\"media_metadata\"])\` against this snapshot. The existing \`are_tilia_data_equal\` call is kept for the other fields (\`timelines_hash\`, \`media_path\`); only the metadata-comparison branch is broken.
- \`on_player_duration_changed\` and \`set_media_duration\` also update the snapshot, since those mutations are system-driven (the player reports the loaded media's length) and shouldn't show up as user edits — without this, the existing \`test_is_file_modified_empty_file\` assertion would flip to True after the test fixture sets the duration to 100.

## Test plan

- New regression tests \`test_is_file_modified_after_notes_edit\` and \`test_is_file_modified_after_title_edit\` cover the case the reporter hit. Both fail without the fix and pass with it.
- All existing modification tests keep passing (\`test_is_file_modified_modified_tile\`, \`..._removed_field_from_media_metadata\`, \`..._modified_media_path\`, \`..._not_modified_after_update\`, \`..._when_modified_timelines\`, \`..._empty_file\`). Their flow already used different objects on each side of the comparison, so the snapshot path is harmlessly redundant for them.
- Targeted suites pass: \`tests/file/\` 20 passed, 1 skipped; \`tests/test_app.py\` 63 passed (1 unrelated sentry-logging test deselected).

## Note for the reviewer

The pre-existing \`media_path\` comparison has the same structural shape — \`on_player_url_changed\` writes the live path to \`self.file.media_path\`, and \`get(Get.MEDIA_PATH)\` reads it from the player, so the two sides drift only because the player doesn't immediately propagate. It happens to *work* in current usage because of that timing, but it's fragile. Out of scope here — happy to follow up if you'd like the same snapshot treatment for it.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -116,12 +116,17 @@ class TiliaState:
     @duration.setter
     def duration(self, value):
         self.app.set_file_media_duration(value)
+        # TiliaState pokes app state directly for test setup; it isn't a
+        # simulated user action, so it shouldn't make is_file_modified()
+        # see an edit (#377).
+        self.file_manager.resync_saved_metadata()
 
     def set_duration(
-        self, value, scale_timelines: Literal["yes", "no", "prompt"] = "prompt"
+        self, value, scale_timelines: Literal["yes", "no", "prompt", "keep"] = "prompt"
     ):
         """Use this if you want to pass scale_timelines."""
         self.app.set_file_media_duration(value, scale_timelines)
+        self.file_manager.resync_saved_metadata()
 
     @property
     def media_path(self):
@@ -253,6 +258,7 @@ def _teardown_youtube_player(player) -> None:
 def tilia(cleanup_requests):
     tilia_ = setup_logic(autosaver=False)
     tilia_.set_file_media_duration(100)
+    tilia_.file_manager.resync_saved_metadata()
     tilia_.reset_undo_manager()
     yield tilia_
     _teardown_youtube_player(tilia_.player)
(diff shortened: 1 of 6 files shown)


azfoo (changes requested):
Added some tests, don't know if their behaviour is expected. Invert their logic if they are!

PR #529, opened 2026-05-12
Speedier pytest

azfoo:
somewhat ironically, this is probably one of the longest pytest runs. this doesn't solve issues with the speed of acquiring the runner image, but the actual speed of running the actual test suite is signifcantly improved.

we might want to consider moving the `pytest-xdist` requirement into the testing group and updating the testing docs to always run tests in parallel.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -60,7 +60,7 @@ jobs:
     - uses: actions/cache@v5
       with:
         path: ${{ matrix.path }}
-        key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
+        key: ${{ runner.os }}-pip-${{ hashFiles('pyproject.toml') }}
         restore-keys: |
          ${{ runner.os }}-pip-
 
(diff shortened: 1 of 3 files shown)


FelipeDefensor (approved):
Edit: this was meant as a review to #530. Ignore and see the review below.

FelipeDefensor (changes requested):
Looks good, but check this point a friend of mine raised:

With pytest-xdist, tests run in worker subprocesses. coverage run -m pytest -n auto only instruments the main pytest
process; the workers (which actually execute the tests) aren't traced unless coverage is configured for subprocess
collection.

To preserve meaningful coverage you need all three:

    In [tool.coverage.run] (pyproject.toml:77): add parallel = true and concurrency = ["multiprocessing"].
    A coverage combine step before coverage report -m.
    Optionally migrate to pytest-cov (which handles this for you via --cov + --cov-context), since pytest-cov already has xdist integration.

Without this, coverage report -m will run successfully but show drastically lower coverage than before. Worth
verifying on a CI run before merge — if numbers look "too low," this is why.

FelipeDefensor (approved):

PR #532, opened 2026-05-13
fix: merge keeps matching label/comments/color instead of joining duplicates

FelipeDefensor:
## Summary
- When every merged hierarchy has the same label or comments, the merged unit adopts that single value instead of concatenating identical copies with the configured separator.
- Same rule for color: when every merged unit shares a non-empty color, the merged unit keeps it (previously color was always dropped on merge).

## Test plan
- [x] `pytest -n auto tests/timelines/hierarchy tests/ui/timelines/hierarchy` — 172 passed.
- [x] New tests cover identical-values-no-separator, mixed empties + identical non-empties, identical colors preserved, differing colors fall back to empty.

### DIFF
--- a/tilia/timelines/hierarchy/timeline.py
+++ b/tilia/timelines/hierarchy/timeline.py
@@ -394,17 +394,23 @@ class HierarchyTLComponentManager(TimelineComponentManager):
         if not success:
             return success, reason
 
-        # preseves label and comments from merged unitst l
+        # Preserve label, comments, and color from merged units. When every
+        # non-empty value for a field matches, keep the single matching value
+        # instead of joining duplicates with the separator.
+        separator = self.timeline.merge_separator
         attr_to_new_value = {}
         for attr in ["label", "comments"]:
-            new_value = hierarchies[0].get_data(attr)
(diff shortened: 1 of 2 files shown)


FelipeDefensor:
Forgot to ask for a review on this one, sorry.

tilia/timelines/hierarchy/timeline.py
```diff
@@ -394,17 +394,23 @@ def is_between_selected_units_and_has_same_parent(h: Hierarchy):
...
+            if not non_empty:
+                continue
+            if len(set(non_empty)) == 1:
+                attr_to_new_value[attr] = non_empty[0]
+            else:
+                attr_to_new_value[attr] = separator.join(non_empty)
+
+        colors = [h.color for h in hierarchies]
+        if colors[0] and all(c == colors[0] for c in colors[1:]):
+            attr_to_new_value["color"] = colors[0]
```
azfoo: perhaps:
```
colors = (h.color for h in hierarchies)
if len(colors) == 1: ...
```

azfoo (changes requested):
just that one line, good to go otherwise.

PR #534, opened 2026-05-14
fix: validate pre-start / post-end length BEFORE add (#495)

### LINKED ISSUE #495: No validation for pre-start and post-end when first created.
It is possible to create a pre-start at `t < media_start` and `t > hierarchy_start` (and v.v. with post-end). Should there be some kind of validation error message? Likewise, when dragging these after creation: should there be some kind of validation error instead of silently deleting them?

azfoo:
closes #495, replaces #522.
- clamps user input to a valid range instead of validating afterwards.
- prevents silent deletes by not allowing the invalid extremity to be created in the first place.


FelipeDefensor (approved):
LGTM. Added tests and handled an edge case.

PR #536, opened 2026-05-14
feat: allow filenames to be passed in as sys.argv[1]

### LINKED ISSUE #466: Can't use TiLiA with "Open with" to open .tla file
In Windows, opening a file with "Open with" makes it so the file is passed to `argv` at index 1. The same happens if an executable is set as the default program to open a certain extension.

Additionally, doing so yields no error message nor anything of the kind, so it's confusing to the user.

azfoo:
closes #466
- Removes `--file` in favour of `file` as a positional argument.
- On a non-existent file or file that does not end with the tilia extension, argparser errors and exits.

### DIFF
--- a/tests/test_boot.py
+++ b/tests/test_boot.py
@@ -1,6 +1,7 @@
 import argparse
 import sys
 from pathlib import Path
+from unittest.mock import MagicMock
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


FelipeDefensor (changes requested):
Looks good, except that there is no error being displayed when opening an invalid file type. TiLiA starts (which is good), but a user that passed a wrong path by mistake is left wondering why fhe file isn't opening.

Btw, feel free to drop the `--file` argument if it is adding too much complexity. I think it is.

FelipeDefensor (approved):
The command-line error message is good, but I would still like to see an error message in the GUI when the file is invalid, although I suppose the user has to try hard to actually attempt to open a non-tla file with "Open with...".

Small detail: update the PR description, it is now stale.

PR #537, opened 2026-05-16
Zoom toolbar

### LINKED ISSUE #470: Creating a new timeline sometimes does not update scrollable vertical area
I am not sure how 

To reproduce:
- Add hierarchy timeline
- Resize window vertically to so that new hierarchy timeline
- Zoom out (this is necessary to reproduce)
- Add another hierarchy timeline

Expected:
- Scrollbar is displayed, as new hierarchy timeline does not fit

Actual:
- New hierarchy timeline is not visible and scrollbar is not displayed

Workaround:
- Zooming updates the scrollbar and makes the timeline visible

azfoo:
- Adds ZoomToolbar with a log-scale slider (1%–1000% visual range), an unbounded spin box
- Zoom is bounded at runtime by _apply_zoom: rejects widths below 1px or above MAX_PLAYBACK_WIDTH; UI reverts to prior zoom on rejection
- Saves zoom ratio per file path in settings and restores it on next open; zoom is re-applied to the timeline width once media duration is known
- Fixes a pre-existing bug where update_height never called scene.setSceneRect, causing the vertical scrollbar to be absent until the user zoomed (closes #470)
- Re-register time_x_converter listeners per test module — cross-module singleton state could corrupt coordinate math under xdist (exposed by the new zoom tests)


FelipeDefensor (approved):
So cool! Thanks :). I debounced the text entry and made the magnifiers a bit smaller. If you are okay with that, feel free to merge.

azfoo:
Looks sensible! 0.6.3 or 0.7.0?

PR #538, opened 2026-05-17
Feat/long operation feedback

### LINKED ISSUE #545: app looks hung when running operations that take a while, clicking anything during the process causes Microsoft to suggest killing the program even though it is still running.
Could implement some kind of visual feedback. Could also look to see if its possible to improve performance; more specifically, try filling beat tl with beats spaced 1s apart on an hour long video without ms suggesting to kill it. (is current performance reasonable?)

azfoo:
**Non-blocking progress feedback for long-running operations**

- New `LongOperationToolbar` (bottom-left) shows a progress bar and label during slow operations.  Stack-based nesting keeps the toolbar visible across multiple concurrent operations and restores the outer label when an inner one finishes.
- `WaitCursor` set for the full duration; `_DialogCursorFilter` lifts it while any `QDialog` is visible and restores it on close. `_InputBlockFilter` swallows mouse clicks while the cursor is active.
- `@long_operation(label)` decorator posts `STARTED`/`DONE` around a function.  The bar is indeterminate by default; posting `PROGRESS(value, maximum)` inside promotes it to determinate. `DONE` is always posted even if the function raises.
- Applied to `App._do_open`, `App._do_load_media`, and `BeatTimeline.fill_with_beats` (with per-beat `PROGRESS` posts).
- Fixes pre-existing bug: `CursorMixIn` was calling `restoreOverrideCursor()` unconditionally on hover-leave, popping the `WaitCursor` off the application stack.  Replaced with `QGraphicsItem.setCursor()`, which does not interact with the override cursor stack.

closes #545

ps. is there anything else that would benefit from this?

### DIFF
--- a/tilia/media/loader.py
+++ b/tilia/media/loader.py
@@ -27,17 +27,27 @@ def load_media(
 
 
 def _change_player_type(player, media_type):
-    player.destroy()
-    # Flush DeferredDelete events so the old widget's C++ object is actually
-    # destroyed before the new player is created. On macOS, QWebEngineView
-    # holds exclusive CoreAudio/AVFoundation resources; if the renderer process
-    # is still alive when QMediaPlayer initialises, they deadlock.
-    # processEvents() alone doesn't flush DeferredDelete (Qt requires a full
-    # event-loop cycle), but sendPostedEvents with DeferredDelete does.
-    from PySide6.QtCore import QCoreApplication, QEvent
+    if player.MEDIA_TYPE == "youtube":
+        # QWebEngineView holds exclusive CoreAudio/AVFoundation resources; if
+        # its renderer process is still alive when QMediaPlayer initialises,
+        # they deadlock on macOS. Destroy the old player and flush
+        # DeferredDelete events so its C++ object is actually gone before the
+        # new player is constructed. processEvents() alone doesn't flush
+        # DeferredDelete (Qt requires a full event-loop cycle), but
+        # sendPostedEvents with DeferredDelete does.
+        player.destroy()
 
-    QCoreApplication.sendPostedEvents(None, QEvent.Type.DeferredDelete)
-    return get(Get.PLAYER_CLASS, media_type)()
+        from PySide6.QtCore import QCoreApplication, QEvent
+
+        QCoreApplication.sendPostedEvents(None, QEvent.Type.DeferredDelete)
+        return get(Get.PLAYER_CLASS, media_type)()
+
+    # Construct the new player before destroying the old one: if the
+    # constructor raises, app.player stays valid. Safe because serve() keeps
+    # _servers_to_requests consistent when ownership changes.
+    new_player = get(Get.PLAYER_CLASS, media_type)()
+    player.destroy()
+    return new_player
 
 
 def get_media_type_from_path(path: str):
--- a/tilia/requests/get.py
+++ b/tilia/requests/get.py
@@ -97,6 +97,15 @@ def serve(replier: Any, request: Get, callback: Callable) -> None:
     Attaches a callback to a request.
     """
 
+    # If a different server already owns this request, remove it from that
+    # server's tracking set so stop_serving_all(old_server) won't later pop
+    # the new server's callback.
+    old_replier, _ = server(request)
+    if old_replier is not None and old_replier is not replier:
+        _servers_to_requests[old_replier].discard(request)
+        if not _servers_to_requests[old_replier]:
+            _servers_to_requests.pop(old_replier)
+
     _requests_to_callbacks[request] = callback
     if replier not in _servers_to_requests.keys():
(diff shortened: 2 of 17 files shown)


FelipeDefensor:
### Performance: `processEvents()` per `PROGRESS` post makes determinate fills superlinear

The decorator is cheap for **indeterminate** ops (`_do_open`, `_do_load_media` post one `STARTED` → one `processEvents`, no per-item cost — those are fine). The problem is **determinate** callers that post `PROGRESS` inside a tight loop: `_on_progress` calls `QApplication.processEvents()` on **every** post, with no throttling. Each pump repaints the growing scene, so `fill_with_beats` turns from a batched loop into a synchronous per-beat repaint.

Micro-benchmark modelling the path (growing `QGraphicsScene`, offscreen — A = one `processEvents` at the end, B = `processEvents` per item, matching `_on_progress`):

| N beats | A (batched) | B (per-beat) | slowdown |
|--:|--:|--:|--:|
| 500 | 0.004 s | 0.11 s | 28× |
| 2000 | 0.018 s | 1.42 s | 81× |
| 4000 | 0.034 s | 5.48 s | 162× |
| 8000 | 0.072 s | 22.4 s | 313× |

Added cost per beat grows with N (0.22 → 2.79 ms) — the signature of ~O(N²): each repaint scales with the items already in the scene. `processEvents()` on an empty queue is ~0.009 ms, so the cost is the forced repaint, not the call itself.

Real-world: `fill_with_beats` counts come from `floor(duration / interval)` or a user-typed amount. ~30 min @ 120 BPM ≈ 3600 beats (~4–5 s freeze); ~60 min ≈ 8000+ (~20 s). Without this feature those fills are <100 ms — so the progress bar ends up being needed *because* the per-beat `processEvents` made the op slow.

**Suggested fix — throttle.** Either post `PROGRESS` every K beats in `fill_with_beats`, or (preferably, so every determinate caller benefits) time-gate inside `_on_progress`: always `setValue`, but call `processEvents()` at most every ~50 ms via a `QElapsedTimer`. Indeterminate ops stay unaffected.

tilia/ui/long_operation.py
```diff
@@ -0,0 +1,124 @@
...
+        if isinstance(obj, QDialog):
+            if event.type() == QEvent.Type.Show:
+                if self._open_count == 0 and QApplication.overrideCursor():
+                    QApplication.restoreOverrideCursor()
+                self._open_count += 1
+            elif event.type() == QEvent.Type.Hide:
+                self._open_count = max(0, self._open_count - 1)
+                if self._open_count == 0 and self._stack:
+                    QApplication.setOverrideCursor(Qt.CursorShape.WaitCursor)
+        return False
```
FelipeDefensor: We can probably unnest some of these if-statements.

tilia/requests/get.py
```diff
@@ -93,6 +93,15 @@ def serve(replier: Any, request: Get, callback: Callable) -> None:
...
 
+    # If a different server already owns this request, remove it from that
+    # server's tracking set so stop_serving_all(old_server) won't later pop
+    # the new server's callback.
+    old_replier, _ = server(request)
+    if old_replier is not None and old_replier is not replier:
+        _servers_to_requests[old_replier].discard(request)
+        if not _servers_to_requests[old_replier]:
+            _servers_to_requests.pop(old_replier)
+
```
FelipeDefensor, line 108: This is a good catch. Two related questions:

- Are there any cases where replacing a server is intentional?
- Do we really want to silently replace instead of raising or returning an error?

tilia/media/loader.py
```diff
@@ -27,8 +27,9 @@ def load_media(
 
 
 def _change_player_type(player, media_type):
+    new_player = get(Get.PLAYER_CLASS, media_type)()
```
FelipeDefensor, line 48: Don't we crash anyway in case the constructor raises, since we are not handling exceptions here? If so, ensuring app.player is still valid does not make a difference.

FelipeDefensor (changes requested):
The core logic looks sound, and the feature is very useful. The main blocker is the performance hit I mentioned in a separate comment, and I am unsure about the server overwrite.

tilia/requests/get.py
azfoo: Yes — this is exactly the case that motivated the fix: `_change_player_type` constructs a new player that takes over as `Get.PLAYER_*` server before the old one calls `stop_serving_all()`. Silent replace is intentional there; raising would break that swap.

tilia/media/loader.py
azfoo: It does matter — _execute_prod catches exceptions at the command layer and shows an error dialog rather than crashing, so if we destroyed-then-constructed and the constructor raised, app.player would be left pointing at a destroyed player after the dialog closes. Construct-before-destroy keeps the app usable after that error.

PR #539, opened 2026-05-18
Feat/player time edit

### LINKED ISSUE #378: Allow for timecode seeking
**Goal**

Move playhead to a precise timecode, e.g. 4:00.1

**Problem**

When moving the playhead, the displayed timecode is not updating, one needs to drop it.

**Proposed solution**

The timecode should always display the playhead's current position, no matter what. For convenience, it would be great to have keyboard shortcuts for moving the cursor by 10s, 1s, 0.1s in either direction.

azfoo:
**Player time display is now a seekable input; playback position restored on file open**

- The current-time label in the player toolbar is now a `QLineEdit`.
Type a time in `HH:MM:SS.f`, `MM:SS.f`, or plain-seconds format and press Enter to seek; the value is clamped to `[0, duration]`.
Pressing Escape reverts without seeking. 
The field is not updated while it has focus so in-progress edits survive playback ticks.
A `QRegularExpressionValidator` allows all intermediate states while typing.
 - Controls are now disabled individually rather than via `setEnabled()` on the whole toolbar, so `time_edit` and the duration label remain enabled in `NO_MEDIA` state when a duration is present (annotation-only files with no media loaded).
 - Current playback position is saved to QSettings on file save and restored as a pending seek on next open. YouTube files are excluded.
 - Adds `parse_media_time()` as the inverse of `format_media_time()`.

(somewhat resolves #378)

### DIFF
--- a/tilia/file/file_manager.py
+++ b/tilia/file/file_manager.py
@@ -350,8 +350,16 @@ class FileManager:
             geometry, window_state = get(Get.WINDOW_GEOMETRY), get(Get.WINDOW_STATE)
             zoom = get(Get.CURRENT_ZOOM)
         except NoReplyToRequest:
+            # No UI is listening (e.g. CLI usage or tests without a full
+            # Qt UI set up), so there's nothing to persist here.
             geometry, window_state, zoom = None, None, None
-        settings.update_recent_files(path, geometry, window_state, zoom)
+        try:
+            current_time = get(Get.MEDIA_CURRENT_TIME)
+        except NoReplyToRequest:
+            current_time = None
+        settings.update_recent_files(
+            path, geometry, window_state, zoom, time=current_time
+        )
 
     def new(self):
         self._setup_file()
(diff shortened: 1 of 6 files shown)


tilia/file/file_manager.py
```diff
@@ -225,7 +225,11 @@ def save(self, data: dict, path: Path | str):
             geometry, window_state = get(Get.WINDOW_GEOMETRY), get(Get.WINDOW_STATE)
         except NoReplyToRequest:
             geometry, window_state = None, None
-        settings.update_recent_files(path, geometry, window_state)
+        try:
+            current_time = get(Get.MEDIA_CURRENT_TIME)
+        except NoReplyToRequest:
+            current_time = None
+        settings.update_recent_files(path, geometry, window_state, time=current_time)
```
FelipeDefensor, line 232: What does this fix? Consider adding comment.

PR #544, opened 2026-05-22
fix: macOS player-swap deadlock, stray timeline view flash, CLI EOF exit

azfoo:
- **`tilia/media/loader.py`** — When swapping player types, explicitly call `QCoreApplication.sendPostedEvents(None, DeferredDelete)`. On macOS deferred deletions do not reliably flush (it can block on platform events first); the direct `sendPostedEvents` call bypasses the platform dispatcher and ensures the old widget — particularly `QWebEngineView` — is truly destroyed before the new player is created, preventing a macOS deadlock.

- **`tilia/ui/timelines/collection/collection.py`** — Embed `TimelineView` into the scene before constructing `TimelineUI`. Prevents an un-parented view from flashing when a timeline's constructor opens a dialog.

- **`tilia/ui/cli/ui.py`** — Treat EOF on stdin as a clean exit (code 0) rather than an error. The CI smoke check now pipes `about` + `quit` into the CLI instead of using a background process, and `QT_DEBUG_PLUGINS` is scoped to the GUI launch only so plugin noise no longer interleaves with CLI output.

### DIFF
--- a/tilia/ui/timelines/collection/collection.py
+++ b/tilia/ui/timelines/collection/collection.py
@@ -496,6 +496,7 @@ class TimelineUIs:
         h = get(Get.TIMELINE, id).get_data("height")
         scene = self.create_timeline_scene(id, w, h)
         view = self.create_timeline_view(scene)
+        view.proxy = self.scene.addWidget(view)
 
         element_manager = ElementManager(timeline_class.ELEMENT_CLASS)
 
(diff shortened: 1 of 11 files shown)


tilia/ui/timelines/collection/collection.py
```diff
@@ -496,6 +496,7 @@ def create_timeline_ui(self, kind: TlKind, id: int) -> TimelineUI:
         h = get(Get.TIMELINE, id).get_data("height")
         scene = self.create_timeline_scene(id, w, h)
         view = self.create_timeline_view(scene)
+        view.proxy = self.scene.addWidget(view)
```
FelipeDefensor, line 499: Is preventing the flashing the only reason why we have the proxy? In any case, we should add a comment explaining its raison d'être.

FelipeDefensor:
LGTM, I just need to test this on my Mac before final approval.

P.S.: *Thank you* for getting rid of the flashing.

FelipeDefensor:
Can't test this thoroughly before solving #548. Will try to do that today.

PR #546, opened 2026-05-25
velopack

azfoo:
## Summary

- Replaces the bare Nuitka standalone zip with Velopack-packaged installers: Setup.pkg (macOS Silicon + Intel), Setup.exe (Windows), AppImage (Linux), plus per-channel full/delta `.nupkg` packages for incremental updates.
- Adds in-app update checking (background thread) with a dialog to download and restart, or pull from git on source installs. Accessible via Help → Check for Updates.
- Wires Velopack lifecycle hooks at startup (`tilia/lifecycle.py`) for install/uninstall and OS file-association registration (.tla files).

## Changes

**`scripts/deploy.py`** — removed sdist-based build path: previously the script created an sdist, extracted it to a temp lib directory, and built from there; now builds directly from the editable install, dropping the `build` package from the build dependency group. Added `_build_velopack()`: runs `vpk pack` after Nuitka, downloads the previous release nupkg for
delta generation, writes `vpk-outdir` and `pack-version` to GITHUB_OUTPUT; `--noPortable` on macOS/Windows only (not supported on Linux); strict semver validation; `_patch_macos_plist()` adds `.tla` associations and returns CFBundleExecutable; Nuitka entry point changed from `tilia/__main__.py` to `tilia/` (package mode).

**`.github/workflows/build.yml`** — .NET 10 via brew on macOS (`setup-dotnet` fails on Intel runners); `vpk` CLI via `dotnet tool`; CI tests run the actual Setup installers instead of extracting zip archives; upload path filters by `pack-version` to exclude old nupkgs from delta generation; release check verifies one `full.nupkg` per channel rather than a fixed count; GH_TOKEN added to build env for `gh release download`.

**`tilia/lifecycle.py`** (new) — handles Velopack install/uninstall hooks and OS file-association wiring (Windows registry, macOS `Info.plist`, Linux MIME/desktop). Called from `boot.py` inside `__compiled__` guard.

**`tilia/updates.py`** (new) — background thread polls for updates via Velopack on compiled builds or `git fetch` on source runs; posts `APP_UPDATE_AVAILABLE` with manager + info payload.

**`tilia/ui/dialogs/update.py`** (new) — update dialog dispatches to Velopack (download + restart) or git (pull + `os.execv`) path.

## To verify
1. Checkout commit <4cc369d> and run the app. A pop-up window should appear to prompt a pull of the branch updates. The app should restart and update to v0.0.2.
2. Download the installer from this [release](https://github.com/azfoo/TiLiA/releases/tag/velopack-1) and install. Similarly, a pop-up window should appear to prompt the download of the next version, and restart and update to v0.0.2. TiLiA should also appear as an "Open with..." option. Uninstalling should clean up all tilia files.

> n.b. I have not tested on Linux and Mac.
> failing tests should be fixed by #544


FelipeDefensor:
Pushed 3 commits to this branch implementing the fixes below (bbd1012a, 8bc79574, b3da5daa):

Built + installed the Windows Velopack package locally and exercised the install, file-association, and update flows. Three defects:

**1. Distribution metadata not embedded → app reports version `0.0.0`, empty repo URL**
`setuptools` `find_namespace_packages` pulled a stray top-level `htmlcov/` (coverage output) into `tilia.egg-info/top_level.txt`. Nuitka then treated `htmlcov` as the owner of the `TiLiA` metadata and skipped embedding it, so `importlib.metadata` resolves nothing at runtime (version → `0.0.0`, `GITHUB_URL` → `""`).
*Fix:* `include = ["tilia*"]` under `[tool.setuptools.packages.find]`, plus `--include-distribution-metadata=TiLiA` in `deploy.py` (reason "user requested" → Nuitka fails loudly instead of silently dropping the metadata).

**2. `.tla` file associations never registered on install**
The Velopack lifecycle hook matched `--velopack-*` argv flags, but Velopack passes `--veloapp-*`. The hook never fired, so no association keys were written.
*Fix:* correct the prefix to `--veloapp-install/-updated/-obsolete/-uninstall/-firstrun` in both the dispatch table and the `_hook_mode` gate.

**3. Updates never offered — every published release is un-updatable** *(main issue)*
"Check for Updates" always reports "up to date", even from an older install. The GitHub releases carry the nupkgs + Setup.exe but **no `releases.<channel>.json`** — the manifest Velopack's GitHub source needs to enumerate versions. Without it, `check_for_updates()` returns `None`.
`vpk pack` *does* emit `releases.<channel>.json` / `assets.<channel>.json` / `RELEASES-<channel>`, but the `upload-artifact` step in `build.yml` globs only `*.nupkg` / `*-Setup.*` / `*.AppImage`, dropping the manifests before the release upload (`files: build/velopack/**`).
*Fix:* add to the artifact `path:`
```
build/*/velopack/releases.*.json
build/*/velopack/assets.*.json
build/*/velopack/RELEASES-*
```
Also: the "Check output" step validates only `*-full.nupkg` + the 4 user-downloads, so CI went green while shipping broken releases — recommend asserting `releases.<channel>.json` per channel there too.

Verified the full path (detect → download → apply → restart) locally by packing a 0.0.3 release and pointing the updater at it; post-update `sq.version` = 0.0.3.

FelipeDefensor:
The 0.0.1 release works on Mac with the same bugs it had on Windows. I imagine the last commits fix it. I will let a local build running and test tomorrow.

FelipeDefensor:
@azfoo, could you fix CI too?

FelipeDefensor (changes requested):
The Mac build worked, after applying the changes I commited. I wasn' t able to run it on Linux. Could you do that after you review my code?

FelipeDefensor:
Is this also ready for review?

azfoo:
not quite. pending a final ci test.

azfoo:
## verify [updates](https://github.com/azfoo/TiLiA/releases/tag/v99.0.0):
- [ ] linux updates
- [x] mac updates
- [x] windows updates

## todo:
- remove [temp] commits

_n.b.:_ failed test run because faked version number exceeds the programmed high version number in test.

FelipeDefensor:
Works great in my Mac!

Edit: I found the issues below later, but the other paths still work perfectly.

FelipeDefensor (commented):
Reviewed the branch (base `ad2a2a2e`) and verified each of the findings below on macOS 15 (Darwin 24.6, arm64) against the installed 99.0.1 build. Two are hard reproductions (the threading crash and the clean-loop failure), one is a live behaviour check (`open -a` on a `.tla`), two are code reads.

tilia/updates.py
```diff
@@ -0,0 +1,193 @@
...
+    """Surface an update-check failure to the user.
+
+    Silent (startup) checks stay quiet — we don't want an error popup on every
+    boot when, e.g., the network is merely unavailable.
+    """
+    if silent:
+        return
+    import tilia.errors
+
+    tilia.errors.display(tilia.errors.VELOPACK_UPDATE_FAILED, message)
```
FelipeDefensor: **Update-check errors crash the app on macOS: `_report` builds the dialog on the background thread.**

`check_for_updates()` runs `_run_check` in a daemon thread. Every failure path in it calls `_report`, which calls `tilia.errors.display` → `post(Post.DISPLAY_ERROR, ...)`. `post()` dispatches listeners synchronously in the calling thread (`tilia/requests/post.py`), so `tilia/ui/dialogs/basic.py::display_error` constructs a `QMessageBox` and calls `box.exec()` off the GUI thread.

Verified on macOS (Darwin 24.6, PySide6 6.9) — constructing and `exec()`-ing that exact dialog from a worker thread terminates the process:

```
QObject::setParent: Cannot set parent, new parent is in a different thread
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException',
    reason: 'NSWindow should only be instantiated on the main thread!'
libc++abi: terminating due to uncaught exception of type NSException
```

Reachable from every non-silent failure: `ImportError` on `velopack`, missing `sq.version`, `UpdateManager(...)` raising, `check_for_updates()` raising (offline), and the `_run_check_safe` backstop. So "Check for Updates" with no network kills the app instead of showing an error.

The success path already handles this correctly — `QtUI.on_update_available` marshals with `QTimer.singleShot(0, self.main_window, ...)`. `_report` needs the same treatment.

tilia/lifecycle.py
```diff
@@ -0,0 +1,299 @@
...
+
+
+def _unregister_windows_file_association() -> None:
+    import winreg
+
+    # Only remove .tla if we own it
+    try:
+        with winreg.OpenKey(winreg.HKEY_CURRENT_USER, _TLA_EXT_KEY) as k:
+            if winreg.QueryValueEx(k, "")[0] == _TLA_PROGID:
+                winreg.DeleteKey(winreg.HKEY_CURRENT_USER, _TLA_EXT_KEY)
```
FelipeDefensor: **Uninstall never removes `Software\\Classes\\.tla`, because the delete order is wrong.**

`_register_windows_file_association` creates both the `.tla` key (default value `TiLiA.tla`) and the `.tla\OpenWithProgids` **subkey**. Here the `.tla` key is deleted first, but `winreg.DeleteKey` fails on a key that still has subkeys, and the `except OSError: pass` swallows it. The second block then only deletes the *value* inside `OpenWithProgids`, leaving the (now empty) subkey behind.

Net effect after install → uninstall: the ProgID and `Applications\TiLiA.exe` keys are gone, but `.tla` survives with its default still pointing at the deleted `TiLiA.tla` ProgID. Double-clicking a `.tla` afterwards gives a broken-handler error instead of falling back to "no association".

Fix: delete `.tla\OpenWithProgids` first, then `.tla`.

tilia/lifecycle.py
```diff
@@ -0,0 +1,299 @@
...
+    except Exception:
+        pass
+
+
+# ---------------------------------------------------------------------------
+# Velopack lifecycle hooks
+# ---------------------------------------------------------------------------
+
+
+def _on_velopack_install(version: str) -> None:
```
FelipeDefensor, line 343: **macOS: the `.tla` association registers, but opening a file from Finder gives a blank document.**

The Info.plist patch in `deploy.py` adds `CFBundleDocumentTypes` / `UTExportedTypeDeclarations`, so TiLiA does show up under "Open with". But macOS does not pass the path in `argv` for a bundled app — LaunchServices sends an `odoc` Apple Event, which Qt turns into a `QFileOpenEvent` on the `QApplication`. Nothing in the codebase handles it: `git grep -n "FileOpen" tilia/` returns nothing, `boot.py` creates a plain `QApplication(sys.argv)` with no `event()` override, and the only path into a file is `args.file` from `setup_parser()`.

Verified live against the installed 99.0.1 build:

```
open -a /Applications/TiLiA.app /tmp/fileopen-probe.tla
```

A new session log appears (`INITIALISED`, blank `SliderTimeline` created) and the probe path never shows up anywhere in it — the app starts empty.

Needs a `QFileOpenEvent` handler on the application object that routes into the same "open file" path as the CLI argument (and it has to buffer the event if it arrives before the UI is built, which on a cold launch it does).

scripts/deploy.py
```diff
@@ -98,97 +137,50 @@ def _clean_version(v: str) -> list[str]:
...
+    icon_path = root / "docs" / "img" / "main_icon.ico"
     exe_args = [
         sys.executable,
         "-m",
         "nuitka",
         f"--output-dir={(outdir / 'exe').as_posix()}",
         f"--product-name={name}",
         f"--file-version={version}",
-        f"--output-filename={out_filename}",
+        f"--output-filename={out_filename if 'mac' in build_os else name}",
```
FelipeDefensor: **macOS: the executable name inside the bundle is version-stamped, so the CLI path changes every release.**

On macOS `--output-filename` is `out_filename` (e.g. `TiLiA-v99.0.1-macos-silicon`) instead of the plain project name used for Windows and Linux. Checked against the installed build:

```
$ /usr/libexec/PlistBuddy -c "Print :CFBundleExecutable" /Applications/TiLiA.app/Contents/Info.plist
TiLiA-v99.0.1-macos-silicon

$ cat /Applications/TiLiA.app/Contents/Resources/sq.version
... <mainExe>Contents/MacOS/TiLiA-v99.0.1-macos-silicon</mainExe>
```

To be clear about what is *not* broken: the bundle itself is stable (`/Applications/TiLiA.app`), because Velopack names it after `--packId`, and updates stay self-consistent since the whole bundle content is replaced.

What does break is the documented terminal entry point — `TiLiA.app/Contents/MacOS/<exe> --user-interface=cli` — whose path changes with every release, so any script, doc or shell alias pointing at it breaks on each update. It also contradicts `_stable_exe_name`'s own docstring ("Stable executable name used by Velopack"); note that on macOS that helper returns the *bundle* name, not an executable, while the real `--mainExe` comes from `_patch_macos_plist`.

Suggest using the plain project name for `--output-filename` on macOS too, and keeping the versioned string only for the uploaded artifact name.

scripts/deploy.py
```diff
@@ -246,48 +240,165 @@ def _build_exe():
...
-            _print([f"\t~{r}"])
+        for r, dirs, files in os.walk(buildlib, False):
+            p = Path(r)
+            _print([f"\t~{p}"])
             for f in files:
-                os.unlink(r / f)
+                os.unlink(p / f)
             for d in dirs:
-                os.rmdir(r / d)
+                os.rmdir(p / d)
```
FelipeDefensor: **The build-folder clean crashes on a second local build, because `os.rmdir` cannot remove a directory symlink.**

`os.walk` does not follow symlinked directories, but it does list them in `dirs`, and `os.rmdir` on a symlink raises `NotADirectoryError`. This is unhandled and sits before the `try`, so `deploy.py` aborts.

It was harmless while the build went through a temp sdist lib dir; now that the macOS `.app` bundle is written into `build/`, the tree is full of directory symlinks. From the installed bundle:

```
Contents/MacOS/PySide6/Qt/lib/QtCore.framework -> ../../../../Frameworks/PySide6/Qt/lib/QtCore.framework
Contents/Frameworks/PySide6/Qt/lib/QtCore.framework/Resources -> Versions/Current/Resources
```

Reduced repro of this exact loop:

```
NotADirectoryError: [Errno 20] Not a directory: '.../build/sub/link'
```

CI runners start clean so this never fires there — it only hits anyone building locally twice. `shutil.rmtree(buildlib)` handles symlinks correctly and replaces the whole hand-rolled walk.

tilia/updates.py
azfoo: failure pathnow uses the same process as success. d2fa9dbc

tilia/lifecycle.py
azfoo: fixed by: a8db846f

scripts/deploy.py
azfoo: stabilised with 638ff6f5. still can't quite use the same as the other os because mac filepaths are case insensitive and plain tilia collides with another dir.

scripts/deploy.py
azfoo: resolved by 2e96f8f1

tilia/lifecycle.py
azfoo: resolved with f644bb77.

tilia/lifecycle.py
azfoo: @FelipeDefensor can you test this to confirm?

azfoo:
Several force pushes to clean up [temp] commits.

Latest builds:
- https://github.com/azfoo/TiLiA/releases/tag/v89.9.8 - https://github.com/azfoo/TiLiA/actions/runs/33514112533
- https://github.com/azfoo/TiLiA/releases/tag/v89.9.9 -  https://github.com/azfoo/TiLiA/actions/runs/33521725840

FelipeDefensor:
Hi Anne,

The update was not working in Linux. I pushed two fixes, feel free to disregard then and write your own.

1. 9b51c999 — fix: locate UpdateNix in AppImage usr/bin
   The Linux branch of _make_locator() built UpdateExePath from current_dir.parents[1], giving <mount>/UpdateNix. vpk actually puts UpdateNix in <mount>/usr/bin/, next to sq.version and the binary. Velopack rejects a locator whose UpdateExePath does not exist, so UpdateManager() raised "This application is not properly installed" and Help > Check for Updates... reported "TiLiA is not properly installed". The silent check at startup failed unnoticed too. Fix: update_exe = current_dir / "UpdateNix".

2. bb25d0c6 — feat: show progress while downloading an update
   Found while testing the first fix. After confirming "Restart and install", the ~270 MB download ran on a daemon thread with no visible feedback. I assumed nothing had happened, closed TiLiA, and that silently killed the download at about 50%. There is now an application-modal progress dialog fed by Velopack's download progress callback; its label switches to "Restarting TiLiA..." and it closes before any error is reported. No cancel button, since Velopack downloads cannot be cancelled.

Tests: test_make_locator_linux_uses_updatenix_and_appimage_env now asserts the exact usr/bin/UpdateNix path instead of endswith(). New tests/test_update_dialog.py covers the progress path and the failure path.

tests/test_update_dialog.py
```diff
@@ -0,0 +1,68 @@
...
+
+def _progress_dialogs() -> list[QProgressDialog]:
+    return [w for w in QApplication.topLevelWidgets() if isinstance(w, QProgressDialog)]
+
+
+def test_velopack_update_shows_download_progress(qtui, monkeypatch):
+    manager = MagicMock()
+    manager.download_updates.side_effect = lambda _update, callback: callback(42)
+    update = MagicMock()
+    update.TargetFullRelease.Version = "89.9.9"
```
FelipeDefensor: Will this break on future versions?

tests/test_update_dialog.py
```diff
@@ -0,0 +1,68 @@
...
+        for dialog in dialogs:
+            dialog.close()
+            dialog.deleteLater()
+
+
+def test_velopack_update_closes_progress_on_failure(qtui, monkeypatch):
+    manager = MagicMock()
+    manager.download_updates.side_effect = RuntimeError("network down")
+    update = MagicMock()
+    update.TargetFullRelease.Version = "89.9.9"
```
FelipeDefensor: Idem.

tests/test_update_dialog.py
```diff
@@ -0,0 +1,68 @@
...
+    monkeypatch.setattr(update_dialog.threading, "Thread", _SyncThread)
+    shown_errors = []
+    monkeypatch.setattr(
+        update_dialog.tilia.errors, "display", lambda *args: shown_errors.append(args)
+    )
+
+    with Serve(Get.FROM_USER_YES_OR_NO, True):
```
FelipeDefensor: There is a newer pattern to be used. See our testing guidelines. Also open an issue to replace all `Serve` with the new pattern. Agents always get this wrong.

FelipeDefensor (approved):
For my part, we are good to go!

I made a few comments for Claude, should be fixed in a few minutes.

tests/test_update_dialog.py
FelipeDefensor: No — the string is only rendered as the progress dialog label and never compared against the app version, so no real version can outgrow it. Made that explicit in 4c0e9201: it is now a named `TARGET_VERSION` constant with a comment, and the test asserts the dialog actually renders it while downloading (previously it was never checked). That assertion also caught that by the end of the flow the label has already switched to "Restarting TiLiA...", so the test now covers both states.

tests/test_update_dialog.py
FelipeDefensor: Same answer, and both tests now share the setup via a `_fake_update()` helper, so the version lives in one place.

tests/test_update_dialog.py
FelipeDefensor: Switched to `patch_yes_or_no_dialog` from `tests.mock` in 4c0e9201, per TESTING.md line 19 (patching the modal covers more code than serving the request). Shared setup also moved into helpers. Opened #605 for migrating the remaining `Serve` call sites: 171 across 20 files, plus 2 `ServeSequence`. Note that int and float prompts have no helper yet, so that part needs new ones.

PR #550, opened 2026-06-22
feat: version-aware .tla loading with migrations

FelipeDefensor:
## Summary

Adds version handling when opening `.tla` files. Until now `open_tla` stored the
file `version` but never read it: opening a file written by a newer TiLiA silently
partial-loaded and could drop unknown data on save, and older-format upgrades were
done with ad-hoc shims scattered inside `deserialize_timelines`.

This PR:

- **Warns + asks confirmation** before opening a file from a newer TiLiA version
  (`open_tla` → `Get.FROM_USER_YES_OR_NO`). Declining aborts the load.
- Adds `tilia/file/migration.py`: an ordered list of small, composable
  `dict -> dict` migrations applied to the raw file data on load.
- Moves the inline `display_position -> ordinal` shim out of
  `deserialize_timelines` into the registry (target `0.1.1`), giving version
  upgrades a single source of truth.
- Registers the upcoming v0.7 timeline-kind rename
  (`MARKER_TIMELINE -> Marker`) as a **dormant** migration (target `0.7.0`)

## Follow-up

When syncing onto the `v0.7.0` branch, delete the inline `_get_timeline_class_string`
shim there — the dormant `0.7.0` migration replaces it.

### DIFF
--- a/tests/file/test_migration.py
+++ b/tests/file/test_migration.py
@@ -22,6 +22,16 @@ def make_timeline(kind, **extra):
     }
 
 
+def make_pre_0_1_1_timeline(kind, display_position, **extra):
+    """A timeline dict as written before the 0.1.1 ordinal migration.
+
+    Pre-0.1.1 files have ``display_position``, not ``ordinal``.
+    """
+    timeline = make_timeline(kind, display_position=display_position, **extra)
+    del timeline["ordinal"]
+    return timeline
+
+
 def make_tla(version, timelines=None):
     return {
         "file_path": "",
@@ -70,9 +80,8 @@ class TestMigrate:
     def test_display_position_to_ordinal(self):
         data = make_tla(
             "0.1.0",
-            {"0": make_timeline("MARKER_TIMELINE", display_position=3)},
+            {"0": make_pre_0_1_1_timeline("MARKER_TIMELINE", display_position=3)},
         )
-        del data["timelines"]["0"]["ordinal"]
 
         migrated, applied = migrate(data, app_version="0.6.2")
 
@@ -102,9 +111,8 @@ class TestMigrate:
     def test_chaining_runs_all_applicable_migrations(self):
         data = make_tla(
             "0.1.0",
-            {"0": make_timeline("HIERARCHY_TIMELINE", display_position=0)},
+            {"0": make_pre_0_1_1_timeline("HIERARCHY_TIMELINE", display_position=0)},
         )
-        del data["timelines"]["0"]["ordinal"]
 
         migrated, applied = migrate(data, app_version="0.7.0")
 
--- a/tests/file/test_migration.py
+++ b/tests/file/test_migration.py
@@ -1,0 +1,159 @@
+import json
+
+from tests.mock import Serve
+from tilia.file.file_manager import open_tla
+from tilia.file.migration import (
+    _parse_version,
+    is_from_newer_version,
+    migrate,
+)
+from tilia.requests import Get
+
+
+def make_timeline(kind, **extra):
+    return {
+        "height": 220,
+        "is_visible": True,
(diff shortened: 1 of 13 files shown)


azfoo:
is this meant to merge into dev instead of main?

FelipeDefensor:
> is this meant to merge into dev instead of main?

Yes. I wonder if we are doing the main vs. dev thing in a unconventional way, Claude always seems to get it wrong.

azfoo:
> > is this meant to merge into dev instead of main?
> 
> Yes. I wonder if we are doing the main vs. dev thing in a unconventional way, Claude always seems to get it wrong.

I suppose there is also main vs stable...

tilia/file/migration.py
```diff
@@ -0,0 +1,89 @@
...
+    for segment in version.split("."):
+        digits = ""
+        for char in segment:
+            if not char.isdigit():
+                break
+            digits += char
+        if not digits:
+            break
+        parts.append(int(digits))
+    return tuple(parts) or (0,)
```
azfoo, line 37: stdlib's platform._comparable_version?

tilia/file/file_manager.py
```diff
@@ -11,11 +12,20 @@
...
 from tilia.requests import Get, Post, get, listen, post, serve
 from tilia.settings import settings
 from tilia.ui import commands
 
+OPEN_FILE_NEWER_VERSION_TITLE = "Open file from a newer version?"
+OPEN_FILE_NEWER_VERSION_PROMPT = (
+    "This file was created with TiLiA {} but you are running {}.\n"
+    "It may not load correctly, and saving could discard data that this "
+    "version doesn't understand.\n\n"
+    "Open it anyway?"
```
azfoo, line 31: this message seems a bit tame. something about "Errors may occur and information may be lost" would probably be more accurate. do we need to add info on where a newer version can be downloaded from?

tilia/file/migration.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def is_from_newer_version(data: dict, app_version: str = VERSION) -> bool:
+    return _parse_version(data.get("version", "0.0.0")) > _parse_version(app_version)
```
azfoo, line 41: should there be a default? or should it error when not found?

(rather this actually depends on whether or not we validate or migrate first...)

tilia/file/file_manager.py
azfoo, line 87: are we sure we want to validate before migrating?

tests/file/test_migration.py
```diff
@@ -0,0 +1,159 @@
...
+        assert migrated["timelines"]["0"]["kind"] == "Marker"
+        assert applied == ["0.7.0"]
+        assert migrated["version"] == "0.7.0"
+
+    def test_chaining_runs_all_applicable_migrations(self):
+        data = make_tla(
+            "0.1.0",
+            {"0": make_timeline("HIERARCHY_TIMELINE", display_position=0)},
+        )
+        del data["timelines"]["0"]["ordinal"]
```
azfoo: add a comment on the del or make a function that adds for >0.1?

azfoo (changes requested):

tilia/file/file_manager.py
azfoo: related to this: what happens when loading a newer version with a tl type that is not known to the older one? pretty sure nothing loads at all because of it, but do verify!

tilia/file/migration.py
azfoo: probably best not to use private functions.

tilia/file/file_manager.py
azfoo: yes, validation only checks that the top level keys are present before migration.

PR #553, opened 2026-07-02
fix: comments icon no longer blocks dragging hierarchy handles

FelipeDefensor:
The comments indicator on adjacent hierarchy units sat on top of the body drag handles, making those boundaries impossible to drag. Reported on macOS; the glyph is narrower on Windows, so it may not reproduce there.

**Cause:** the comments icon has `z = level + 1`, above the drag handles' `z = 0`. Click dispatch uses `QGraphicsView.itemAt`, which returns the topmost item — and the icon's bounding box overhangs the handle's x, so clicks resolved to the icon, which isn't a drag trigger.

**Fix:** raise the drag handles above the decorative overlays (`HANDLE_Z = 100`) so they always win the hit-test. Handles were already above all bodies, so only handle-vs-icon stacking changes. Also covers the loop icon's matching overlap with the start handle.

Added a regression test asserting the comments icon stacks below both handles.

### DIFF
--- a/tests/ui/timelines/hierarchy/test_hierarchy_ui.py
+++ b/tests/ui/timelines/hierarchy/test_hierarchy_ui.py
@@ -309,6 +309,14 @@ class TestCommentsIndicator:
         hui.set_data("end", 100)
         assert hui.comments_icon.isVisible()
 
+    def test_does_not_stack_above_drag_handles(self, tlui):
+        # Regression: the comments emoji's bounding box overhangs the end
+        # handle's x. If the icon stacks above the handle it wins the itemAt
+        # hit-test and the boundary can't be dragged.
+        hui = self.create_hierarchy_ui(tlui, comments="something")
+        assert hui.comments_icon.zValue() < hui.start_handle.zValue()
+        assert hui.comments_icon.zValue() < hui.end_handle.zValue()
+
 
 class TestDoubleClick:
     def test_posts_seek(self, tlui, tilia_state):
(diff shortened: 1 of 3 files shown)


tests/ui/timelines/hierarchy/test_hierarchy_ui.py
```diff
@@ -309,6 +309,14 @@ def test_is_shown_when_hierarchy_gets_big_enough(self, tlui):
         hui.set_data("end", 100)
         assert hui.comments_icon.isVisible()
 
+    def test_does_not_stack_above_drag_handles(self, tlui):
+        # Regression: the comments emoji's bounding box overhangs the end
+        # handle's x (notably wide on macOS). If the icon stacks above the
```
FelipeDefensor: Don't mention macOS.

tilia/ui/timelines/hierarchy/handles.py
```diff
@@ -6,6 +6,14 @@
 
 from tilia.ui.timelines.cursors import CursorMixIn
 
+# Drag handles must be hit-tested above the decorative overlays (label,
+# comments and loop icons, which sit at z = level + 1 in element.py). Their
+# bounding boxes can overhang a handle's x — e.g. the wide comments emoji on
+# macOS overlaps the end handle — so a lower z would let the overlay intercept
```
FelipeDefensor: HANDLE_Z = 100 seems less excessive. Also make the comment more concise.

FelipeDefensor: Done in 8e3ac315 — `HANDLE_Z = 100`, comment cut to two lines.

tests/ui/timelines/hierarchy/test_hierarchy_ui.py
FelipeDefensor: Done in 8e3ac315 — dropped the macOS mention.

PR #555, opened 2026-07-21
fix: don't scale or crop timelines when opening a saved file (#453)

### LINKED ISSUE #453: Opening files with YouTube videos prompts for scaling unnecessarily
To reproduce, open [PinkFloyd_Echoes_MK - Kopie.tla.json](https://github.com/user-attachments/files/27082644/PinkFloyd_Echoes_MK.-.Kopie.tla.json)

FelipeDefensor:
Closes #453.

## Problem

Opening a `.tla` sets the media duration twice: first from the file's stored `media length`, then again when the player reports the actual duration. For YouTube sources these two values usually differ by a fraction of a second, so TiLiA prompted the user to scale timelines that already matched the saved media.

## Fix

Route the file-open media load through a new `scale_timelines="keep"` mode. On open, the timelines were saved to match this media, so the duration the player reports must not scale, crop, or prompt — it should only update the stored duration.

`"keep"` rather than `"no"`: `"no"` still **crops** when the reported duration comes back *shorter* than the stored value (`App.on_media_duration_changed`), which silently deletes any components past the new end. YouTube's asynchronous duration can be a few milliseconds shorter than the stored value, so `"no"` risked data loss. Explicit **Load Media** actions keep the default `"prompt"`, and the media-not-found → replacement-media path on open also keeps `"prompt"`.

## Test

`test_async_duration_after_open_does_not_prompt_or_change_timelines` — parametrized over a **shorter** and a **longer** async duration reported after open; asserts no scale prompt fires and no component is scaled or cropped. Built with `save_and_reopen` (a real temporary `.tla`). Verified it fails under `"no"` (the end marker is cropped: `assert 0 == 1`) and passes under `"keep"`.

## Test plan

- [x] `pytest tests/test_app.py::TestFileLoad tests/test_app.py::TestScaleCropTimeline tests/test_app.py::TestMediaLoad` (22 passed)
- [x] pre-commit (black, ruff)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -306,6 +306,31 @@ class TestFileLoad:
         assert len(marker_tl) == 1
         assert marker_tl[0].get_data("time") == pytest.approx(marker_time)
 
+    def test_large_async_duration_change_after_open_still_prompts(
+        self, tilia, tilia_state, marker_tlui, tls, tmp_path
+    ):
+        # A duration change larger than DURATION_JITTER_TOLERANCE
+        # (tilia/app.py) is not YouTube's async jitter -- even under "keep"
+        # it must still be treated as a genuine media change and prompt to
+        # scale, unlike the small-jitter case in the test above.
+        tilia_state.media_path = EXAMPLE_MEDIA_PATH
+        tilia_state.set_duration(EXAMPLE_MEDIA_DURATION, scale_timelines="no")
+        marker_time = EXAMPLE_MEDIA_DURATION - 0.5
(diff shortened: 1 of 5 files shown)


FelipeDefensor:
@azfoo, sorry for the dealy on opening, but this one is important. The bug is *very* annoying. Ideally to be landed on the next minor version.

I didn't test the AI fix.

azfoo:
@FelipeDefensor will do, no worries. do also see the last email I sent you.

PR #570, opened 2026-08-21
fix: keep hierarchy paste-complete boundaries exact so grouping doesn't fail

### LINKED ISSUE #396: Valid hierarchy unit grouping fails
Trying to group the units selected in the image fails:

Image

FelipeDefensor:
Grouping components derived from a pasted hierarchy subtree failed with
`Grouping failed: Grouping component would overlap with existing component.`

Closes #396.

## Reproducing

Reported against a real file. In a hierarchy timeline:

1. Copy the top-level hierarchy of one timeline.
2. Paste it into a single hierarchy of another with **Paste complete**.
3. Create a child in one of the pasted components.
4. Decrease that child's level.
5. Split the child in two.
6. Try to group the two halves.

Expected the halves to be grouped; got the overlap error instead.

## Cause

`_create_child_from_paste_data` rescaled each copied component with arithmetic
that differed between its start and its end: the start was taken relative to the
source parent's **start**, the end relative to the source parent's **end**. A time
appearing in two components — as one's end and the next one's start — therefore
went through different operations and came out a few ulps apart. Nesting
compounded this, since every level re-derived its scale factor from the
already-rounded bounds of the level above.

In the reported file the pasted `Modulation` component starts at
`134.18723783376336` while its left-hand neighbours end at `134.1872378337634` —
4e-14 further right, so arithmetically *inside* the group span. Since the
grouping validators compare boundaries with exact float operators, that
sub-ulp gap read as a real overlap.

## Fix

Three commits, one per concern:

**`fix: keep hierarchy paste-complete boundaries exact`** — build a single affine
map at the root of the paste and apply it to the absolute times of the whole
subtree, so the same time maps to the same float everywhere it occurs. This is
mathematically identical to the old per-level nesting (composing affine maps
over nested intervals yields the root map) but without the intermediate
rounding. Endpoints are anchored exactly so the pasted subtree lines up flush
with the target.

**`fix: tolerate float drift when comparing hierarchy boundaries`** — compare
boundaries with a 1 ns tolerance instead of exact operators. Needed
independently of the above: files saved before this fix already contain the
drifted values, and exact comparison is fragile for any rescaling operation.
1 ns sits far above the float noise (~1e-13) and far below any meaningful
musical interval, so genuine overlaps are still rejected.

The tolerance is applied to *every* boundary comparison that runs over stored
geometry, not just the two grouping validators, so the helpers cannot disagree
about whether the same timeline is valid:

- `get_parent` / `get_children` — a parent whose start sits a few ulps after its
  child's start was not found, silently orphaning the child. Everything
  depending on the parent link then misbehaves: delete cascade, level changes,
  copy-with-children, and the `has_same_parent` check inside grouping itself.
- `get_boundary_conflicts` — reported the very grouping unit the validators had
  just accepted as conflicting with its drifted neighbour.

Each comparison keeps its original strictness; only the equality test changes.

Checked `scale_segmentlike` (media-duration changes) for the same class of bug —
it applies `t * factor` uniformly to starts and ends, so shared boundaries stay
identical. Paste-complete was the only offender.

**`fix: report paste-complete child creation failures to the user`** — a child of
the pasted subtree that could not be created was skipped with only a log line,
so the user saw a silently incomplete paste. It now goes through the same
`COMPONENTS_PASTE_ERROR` dialog as the other paste-complete failures. Pasting
stays best-effort — a failed child is skipped along with its own subtree and the
remaining siblings are still pasted — so the reasons are collected across the
whole operation and reported once at the end rather than one dialog per failure.

Also folded into the paste commit: the unused `children_of_element` accumulator
is gone, and a failed `create_component` no longer flows into
`get_component_ui(None)`.

## Tests

Each new test was confirmed to fail without the corresponding fix and pass with it.

- Three tests asserting exact float equality of shared boundaries after
  paste-complete — for siblings, for grandchildren (where the old error
  compounded), and for an identity rescale. Two of them fail on `dev` with the
  visible drift `3.8999999999999995` vs `3.9000000000000004`.
- An end-to-end test walking the whole reported flow through commands
  (paste-complete → create_child → decrease_level → split → group), which fails
  on `dev` with the exact reported error message.
- Two backend tests for the grouping tolerance: one where a neighbour's end sits
  a few ulps inside the group span (must succeed), one where it sits well inside
  (must still be rejected).
- Three more for genealogy and conflict detection: a parent whose start drifted
  past its child's start, one whose end drifted before its child's end, and
  coincident-but-drifted boundaries that must not be reported as conflicting.
- Guard tests against the tolerance being too coarse: a genuinely too-narrow
  parent is still not adopted, an overlap well past a boundary is still
  reported, and a drifted duplicate on the same level is still a conflict.
- Two tests for the paste-complete failure dialog: it fires with the creation
  failure's reason, and it stays silent when every child is created.

Every commit is green on its own. Full suite: 1844 passed, 13 skipped.
`pre-commit` (black + ruff) clean.

### DIFF
--- a/tests/ui/timelines/hierarchy/test_hierarchy_timeline_ui.py
+++ b/tests/ui/timelines/hierarchy/test_hierarchy_timeline_ui.py
@@ -17,6 +17,11 @@ def tlui(hierarchy_tlui):
     return hierarchy_tlui
 
 
+def get_hierarchy(tlui, start: float, level: int) -> Hierarchy:
+    """Look a hierarchy up by its data, for setups that aren't in sorted order."""
+    return next(c for c in tlui.timeline if c.start == start and c.level == level)
+
+
 def set_dummy_copy_attributes(hierarchy: Hierarchy) -> None:
     for attr in HierarchyUI.DEFAULT_COPY_ATTRIBUTES.values:
         if attr == "color":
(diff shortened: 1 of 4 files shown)


tests/ui/timelines/hierarchy/test_hierarchy_timeline_ui.py
```diff
@@ -471,6 +476,137 @@ def test_paste_with_children_into_different_level_fails(self, tlui):
...
+        assert right.end == 100
+
+    def test_group_split_child_after_paste_complete_with_boundary_drift(self, tlui):
+        # Regression for the reported bug: create a child in a pasted unit,
+        # lower its level, split it and group the halves. This used to report
+        # "Grouping component would overlap with existing component" because
+        # paste-complete left the pasted siblings' shared boundary a few ulps
+        # apart. Kept as an end-to-end guard over the whole reported flow.
+        # Source subtree lives at [40, 70]; the two level-3 children share the
+        # boundary 45, pasted into the shorter target [0, 23.4].
```
FelipeDefensor: The bug will not exist anymore, so this should be reworded.

tests/ui/timelines/hierarchy/test_hierarchy_timeline_ui.py
```diff
@@ -471,6 +476,137 @@ def test_paste_with_children_into_different_level_fails(self, tlui):
...
+        root = get_hierarchy(tlui, 40, 4)
+        target = get_hierarchy(tlui, 0, 4)
+
+        tlui.select_element(tlui.get_element(root.id))
+        commands.execute("timeline.component.copy")
+        tlui.deselect_all_elements()
+        tlui.select_element(tlui.get_element(target.id))
+        commands.execute("timeline.component.paste_complete")
+
+        # the pasted right child, still childless (the "Modulation" analogue).
```
FelipeDefensor: "Modulation" shouldn't be mentioned expressely. The comment shouldn't depend on the issue that sparked the fix.

tilia/ui/timelines/hierarchy/timeline.py
```diff
@@ -181,60 +184,83 @@ def paste_single_into_selected_elements(self, paste_data: list[dict]):
...
 
-                if child_paste_data.get("children", None):
-                    self.paste_with_children_into_element(
-                        child_paste_data, self.get_component_ui(child_component)
-                    )
+            if child_component is None:
+                logger.error(
+                    "Could not paste hierarchy child into "
+                    f"'{element.tl_component}': component creation failed."
+                )
```
FelipeDefensor: This error should be displayed to the user using the usual methods.

tilia/timelines/hierarchy/timeline.py
```diff
@@ -17,6 +18,27 @@
 from ..base.timeline import Timeline, TimelineComponentManager, TimelineFlag
 from .components import Hierarchy
 
+# Operations that rescale components (e.g. paste-complete) recompute each
+# component's boundaries independently, so times that were originally
+# coincident can drift apart by a few ulps (~1e-13). Compare hierarchy
+# boundaries with a tolerance far below any meaningful interval so this
+# float noise is not mistaken for an overlap.
```
FelipeDefensor: Mention this is now fixed and that the behaviour is a safeguard for legacy files.

tests/ui/timelines/hierarchy/test_hierarchy_timeline_ui.py
FelipeDefensor: Reworded, and the test renamed `test_group_split_child_after_paste_complete_with_boundary_drift` → `test_group_split_child_of_pasted_hierarchy`. The comment now states what the test guarantees — an end-to-end guard over an editing flow that runs entirely on rescaled boundaries — instead of quoting the error it used to produce. Did the same to the sibling and grandchild test comments, which had the same past-tense framing.

tests/ui/timelines/hierarchy/test_hierarchy_timeline_ui.py
FelipeDefensor: Removed, including the local variable, which is now `pasted_right_child`. No occurrences left in the file. The comment now just explains why that child is the one being split (it is left childless in the fixture).

tilia/ui/timelines/hierarchy/timeline.py
FelipeDefensor: Now goes through `_display_paste_complete_error` / `COMPONENTS_PASTE_ERROR`, the same path as the other paste-complete failures, and `on_paste_element_complete` returns `False`. `_create_child_from_paste_data` returns `(component, reason)` and the reasons bubble up through `_paste_subtree_into_element` → `paste_with_children_into_element`.

Reported once at the end rather than per failure: pasting is best-effort (a failed child is skipped along with its own subtree, remaining siblings still paste), so a broken subtree would otherwise stack one modal dialog per failed child. Covered by two tests — one asserting the dialog fires with the failure reason, one asserting it stays silent on a clean paste.

tilia/timelines/hierarchy/timeline.py
FelipeDefensor: Rewritten. It now says rescaling no longer introduces drift — paste-complete maps the whole subtree with a single affine map, and `scale_segmentlike` applies one factor uniformly — and that the tolerance exists as a safeguard for files saved before that fix.

While reworking it I found the tolerance was only half applied: `get_parent`, `get_children` and `get_boundary_conflicts` still compared exactly, on the same drifted data. A parent whose start sat a few ulps after its child's start was not found by `get_parent`, silently orphaning the child, and `get_boundary_conflicts` reported the very grouping unit the validators had just accepted as conflicting with its drifted neighbour. All three now route through the same `_le` / `_lt` / `_times_close` helpers, with guard tests for the tolerance being too coarse. Folded into that commit, which is now `fix: tolerate float drift when comparing hierarchy boundaries`.

tilia/timelines/hierarchy/timeline.py
```diff
@@ -17,6 +18,31 @@
...
+
+def _le(a: float, b: float) -> bool:
+    """``a <= b``, treating near-coincident times as equal."""
+    return a < b or _times_close(a, b)
+
+
+def _lt(a: float, b: float) -> bool:
+    """``a < b``, treating near-coincident times as equal (i.e. not less)."""
+    return a < b and not _times_close(a, b)
+
```
azfoo, line 45: is it worth converting each start/end into its own object with dedicated dunder methods? this seems to be a comparison that might be useful across all the different tls.

azfoo (changes requested):

PR #584, opened 2026-08-25
fix: close the score viewer when the score timeline is cleared

### LINKED ISSUE #472: Clearing a score timeline does not close the SVG viewer

FelipeDefensor:
Alternative to #481, rebuilt on current `dev`. Same diagnosis, deeper fix.

Closes #472.

## The bug is real and reachable

Clearing a score timeline and then letting playback run for one tick reopens the viewer with the score that was just discarded: `on_audio_time_change` reads `ScoreTimelineUI.svg_view`, whose getter *constructs* an `SvgViewer` when none is registered, loads `svg_data` into it and shows it. Instrumenting `SvgViewer.show` confirms one call per clear-plus-tick on `dev`. The score also survived into the saved file, so reopening a cleared timeline brought it back.

## Why not #481 as it stands

#481 guards `reset_svg` alone. That leaves the auto-creating property in place at five other call sites, and it introduces a state (`svg_data == ""` on a live timeline) that the rest of the code cannot handle:

- **Undo/redo crashes.** With `svg_data` blanked on clear, restoring a cleared state pushes `""` through the update trigger into `SvgViewer.load_svg_data`, which hands it to `etree.fromstring` — `XMLSyntaxError: Document is empty`.
- **The new guard catches the wrong thing.** Only one callback can be attached per `Get` member, so with two score timelines the viewer of one answers for the other's id and returns `None` rather than raising. `except NoReplyToRequest` does not cover that; `reset_svg` raises `AttributeError`. `on_audio_time_change` had the same latent crash.
- **The viewer still comes back.** `reset_svg` deletes it, and the next playback tick rebuilds it through the property. It stays invisible only because `svg_data` happens to be empty by then, so the two halves of that PR are load-bearing together rather than independently correct.

## What this does instead

**`528eb264`** — `load_svg_data` treated only `success=False` from `set_data` as a failure, but `viewer_beat_x` uses `validate_pre_validated`, so the call always succeeds and `SCORE_SVG_CREATE_ERROR` was unreachable. A render without usable beat markers silently produced an empty mapping. Check the returned mapping too.

**`6d391c06`** — `svg_view` no longer constructs anything; it returns `None` when there is no viewer, and the explicit `get_or_create_svg_view()` handles creation. All six read sites updated, plus `ScoreAnnotationUI.svg_view`, which keeps the creating behaviour it relies on. This is what actually stops the resurrection, and it fixes the `None`-return crash at the same time.

**`9a13afb9`** — the fix proper. `ScoreTLComponentManager.clear` discards `svg_data` (after `super().clear()`, so the deletion loop still sees the score it is clearing), and `update_svg_data` closes the viewer instead of loading when there is no score left. `load_svg_data` refuses empty data outright.

**`5dd1a7fc`** — docs; see below.

**`b106beac`** — `_setup_svg_view` now goes through `get_or_create_svg_view`, which sets the measure tracker's visibility from the viewer instead of calling `show()` unconditionally. Test pins the behaviour that must be preserved.

## Tests

16 added, covering: the viewer closing on clear and staying closed through playback, `svg_data` being discarded, restoring a cleared state, undo/redo, save-and-reload, `reset_svg` with no viewer and with another timeline owning the slot, and the revived beat-position error. Each fails without its corresponding change.

Full suite: 1847 passed, 0 failed.

## Two pre-existing bugs found on the way, left for their own PRs

- **`viewer_beat_x` is never cleared**, so a score imported into a timeline that already had one is laid out against the previous score's beat positions. Clearing it here is not safe yet: `_restore_timeline_state` restores `svg_data` before `viewer_beat_x`, so an undo would reload the already-marker-stripped SVG while the mapping is empty and recomputation yields nothing.
- **`components_hash` is nondeterministic across undo.** `TimelineComponentManager.restore_state` recreates components in set-iteration order while `hash_components()` hashes them in list order, so a restored state's hash depends on `PYTHONHASHSEED` (fails on seeds 0, 1, 4; passes on 2, 3, 5 — same on clean `dev`). That makes `undoable()` unusable on any timeline with several components and can mark an unmodified file as modified. The undo/redo test here asserts restored values directly and carries a comment pointing at this.

## Unrelated: parallel test runs

`pytest-xdist` was already a dev dependency but nothing documented it, so the documented way to run the suite takes about eight minutes instead of one and a half. `5dd1a7fc` adds `pytest -n auto` to CLAUDE.md, along with the two intermittent-failure sources that look like regressions but are not — cross-test interference under `-n auto`, and the `PYTHONHASHSEED` dependency above.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -23,7 +23,8 @@ tilia                       # Qt GUI
 tilia --user-interface cli  # CLI (source-only, not in compiled builds)
 
 # Tests (pytest-env auto-sets ENVIRONMENT=test and QT_QPA_PLATFORM=offscreen)
-pytest                                                  # full suite
+pytest                                                  # full suite (~10 min)
+pytest -n auto                                          # full suite in parallel (~1.5 min, pytest-xdist)
 pytest tests/ui/timelines/marker/test_marker_timeline_ui.py    # one file
 pytest tests/ui/timelines/marker/test_marker_timeline_ui.py::TestCreateDelete::test_create_at_selected_time  # one test
 pytest -k marker                                        # by keyword
@@ -39,6 +40,14 @@ python scripts/deploy.py <ref_name> <os_type>  # output in build/<os_type>/exe
 ```
 
 Default pytest timeout is 10s (pytest-timeout). `ffmpeg` must be on PATH for audio export/convert features.
+
+Prefer `-n auto` — the suite takes minutes serially and about a minute in parallel. A handful of
+tests (notably in `tests/test_app.py` and the CLI roundtrip tests) fail intermittently under
+`-n auto`, and which ones fail varies between runs; this predates any given change, so confirm a
+failure reproduces serially before investigating it. The same goes for `undoable()` on a timeline
+with several components: `TimelineComponentManager.restore_state` recreates components in
+set-iteration order while `hash_components()` hashes them in list order, so whether the assertion
+passes depends on `PYTHONHASHSEED`.
 
 ## Architecture
 
--- a/tests/ui/timelines/score/test_score_timeline_ui.py
+++ b/tests/ui/timelines/score/test_score_timeline_ui.py
@@ -375,3 +375,80 @@ class TestAudioTimeChange:
         post(Post.PLAYER_CURRENT_TIME_CHANGED, 1.0, None)
 
         assert score_tlui.svg_view is None
+
+
+class TestClear:
+    @staticmethod
+    def _clear(score_tlui):
+        with Serve(Get.FROM_USER_YES_OR_NO, True):
+            commands.execute("timeline.clear", score_tlui)
+
+    def test_closes_the_svg_viewer(self, score_tlui, note, tls):
+        tls.set_timeline_data(score_tlui.id, "svg_data", SVG_WITH_MARKERS)
+
+        self._clear(score_tlui)
+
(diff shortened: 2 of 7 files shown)


CLAUDE.md
```diff
@@ -40,6 +41,14 @@ python scripts/deploy.py <ref_name> <os_type>  # output in build/<os_type>/exe
 
 Default pytest timeout is 10s (pytest-timeout). `ffmpeg` must be on PATH for audio export/convert features.
 
+Prefer `-n auto` — the suite takes minutes serially and about a minute in parallel. A handful of
+tests (notably in `tests/test_app.py` and the CLI roundtrip tests) fail intermittently under
+`-n auto`, and which ones fail varies between runs; this predates any given change, so confirm a
+failure reproduces serially before investigating it. The same goes for `undoable()` on a timeline
+with several components: `TimelineComponentManager.restore_state` recreates components in
+set-iteration order while `hash_components()` hashes them in list order, so whether the assertion
+passes depends on `PYTHONHASHSEED`.
```
azfoo, line 50: though also good to check which tests are leaking data / affecting the other tests. could potentially indicate a real world user interaction sequence issue.

